PR #23, opened 2022-12-09
Surface revert reason even if one cannot find the handle in the bytecode

nambrot:


yorhodes (approved):

src/features/debugger/debugMessage.ts
```diff
@@ -236,6 +236,17 @@ async function checkMessage(
   } catch (err: any) {
     logger.info('Estimate gas call failed');
 
+    const errorReason = extractReasonString(err);
+    logger.debug(errorReason);
+
+    if (errorReason.includes(`execution reverted:`)) {
+      return {
```
jmrossy, line 243: Wouldn't this cause us to miss out on more specific reasons for the bytecode and ICA cases just below

jmrossy (changes requested):

src/features/debugger/debugMessage.ts
yorhodes: yeah great catch
probably move this into the no handle block?

yorhodes (changes requested):

jmrossy (approved):

yorhodes (approved):

PR #34, opened 2023-04-15
PI UX improvements

jmrossy:
- [x] Update to SDK 1.3.2
- [x] Add support for chain metadata in the URL
- [x] Add support for PI message search on message page
- [x] Improve missing chain metadata error
- [x] Improve chain data input UX
- [x] Update warp template to use new query param

Based on feedback from: https://github.com/hyperlane-xyz/issues/issues/423#issuecomment-1503491563

### DIFF
--- a/src/features/chains/ConfigureChains.tsx
+++ b/src/features/chains/ConfigureChains.tsx
@@ -1,6 +1,6 @@
 import { ChangeEventHandler, useState } from 'react';
 
-import { mainnetChainsMetadata, testnetChainsMetadata } from '@hyperlane-xyz/sdk';
+import { ChainName, mainnetChainsMetadata, testnetChainsMetadata } from '@hyperlane-xyz/sdk';
 
 import { SolidButton } from '../../components/buttons/SolidButton';
 import { XIconButton } from '../../components/buttons/XIconButton';
@@ -8,6 +8,7 @@ import { ChainLogo } from '../../components/icons/ChainLogo';
 import { Card } from '../../components/layout/Card';
 import { Modal } from '../../components/layout/Modal';
 import { links } from '../../consts/links';
+import { useMultiProvider } from '../../multiProvider';
 
 import { tryParseChainConfig } from './chainConfig';
 import { useChainConfigs } from './useChainConfigs';
@@ -15,6 +16,7 @@ import { getChainDisplayName } from './utils';
 
 export function ConfigureChains() {
   const { chainConfigs, setChainConfigs } = useChainConfigs();
+  const multiProvider = useMultiProvider();
 
   const [showAddChainModal, setShowAddChainModal] = useState(false);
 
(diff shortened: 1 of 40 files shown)


src/features/chains/useChainConfigs.ts
```diff
@@ -0,0 +1,63 @@
...
+
+// Use the chainConfigs from the store but with any
+// chainConfigs from the query string merged in
+export function useChainConfigsWithQueryParams() {
+  const { chainConfigs: storeConfigs, setChainConfigs } = useChainConfigs();
+  const queryVal = useQueryParam(CHAIN_CONFIGS_KEY);
+
+  useEffect(() => {
+    if (!queryVal) return;
+    const decoded = fromBase64<ChainMetadata[]>(queryVal);
```
nambrot, line 33: Is there not some out of the box way to encode the chain config as plaintext query params?

src/features/chains/utils.ts
```diff
@@ -1,19 +1,20 @@
+import { MultiProvider } from '@hyperlane-xyz/sdk';
+
 import { Environment } from '../../consts/environments';
-import { getMultiProvider } from '../../multiProvider';
 import { toTitleCase } from '../../utils/string';
 
-export function getChainName(chainId?: number) {
-  return getMultiProvider().tryGetChainName(chainId || 0) || undefined;
+export function getChainName(mp: MultiProvider, chainId?: number) {
```
nambrot, line 6: Nit: I guess multiProvider is a handful to write every time but it might be nice to be consistent

src/features/deliveryStatus/fetchDeliveryStatus.ts
```diff
@@ -24,12 +24,13 @@ import {
...
 export async function fetchDeliveryStatus(
+  multiProvider: MultiProvider,
   message: Message,
-  multiProvider = getMultiProvider(),
 ): Promise<MessageDeliveryStatusResponse> {
   const destName = multiProvider.getChainName(message.destinationChainId);
+  const destEnv = getChainEnvironment(multiProvider, destName);
   // TODO PI support here
-  const destMailboxAddr = hyperlaneCoreAddresses[destName]?.mailbox;
+  const destMailboxAddr = hyperlaneEnvironments[destEnv][destName]?.mailbox;
```
nambrot, line 33: Looking at this code, should we even separate by `environments`? It seems like more and more like a relict of v1?

nambrot (approved):
I have to admit I mostly skimmed the PR. I'm curious on how you feel about the complexity of the logic in the explorer now and whether you would want some time to refactor before making some of the other PI changes for debugging and gas estimations, as well as further down the line things like paying for gas and self-relaying by fetching ISM metadata

src/features/deliveryStatus/fetchDeliveryStatus.ts
jmrossy: We shouldn't, this is because asa removed the hyperlaneCoreAddresses object in the SDK. I plan to add it back soon

src/features/chains/useChainConfigs.ts
jmrossy: The nested structure makes it difficult. It was this or add a 3rd part lib like `qs` everywhere we want to serialize it.

src/features/chains/ConfigureChains.tsx
```diff
@@ -178,7 +189,9 @@ const customChainTextareaPlaceholder = `{
   } ],
   "blocks": { "confirmations": 1, "estimateBlockTime": 13 },
   "contracts": {
-    "mailbox": "0x123..."
+    "mailbox": "0x123...",
+    "multisigIsm": "0x123...",
```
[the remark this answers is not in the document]
jmrossy: Eventually it would be for better gas estimation. I'll hide it for now

PR #38, opened 2023-05-04
PI delivery checking & debugging support

jmrossy:
- Add PI support for msg debugging and delivery status fetching
- Ensure chain config syncing only runs once per page
- Move MultiProvider into store to avoid frequent re-construction
- Remove old TxDebugger page and cleanup related code
- Improve error reason surfacing from SmartProvider
- Update widgets to 1.3.4 for timeline fix

Fixes https://github.com/hyperlane-xyz/hyperlane-explorer/issues/33


src/features/providers/SmartMultiProvider.ts
```diff
@@ -0,0 +1,35 @@
+import {
```
jmrossy, line 1: Not new code, moved from multiprovider.ts to avoid import cycle

PR #62, opened 2024-03-12
feat: use SmartProvider from SDK

paulbalaji:
- use SmartProvider from SDK
- drive-by removal of goerli mentions

resolves https://github.com/hyperlane-xyz/issues/issues/991


jmrossy (approved):
Thank you!

PR #67, opened 2024-04-24
fix: replace use tooltip instead of toast for show tip information

tiendn:
We should use tooltips to display tip information instead of Toasts.

**Reasons:**

- Users have to shift their focus from the current help area to the bottom right corner, resulting in poor user experience. It becomes unclear which help was hovered over after moving the mouse.
- Toasts have a delay that is too long and are limited to a maximum of two. Even when moving to another location or page, they still appear and continue to display if clicked multiple times.

**Solution:**
- Implement tooltips using [react-tooltip](https://react-tooltip.com/docs/getting-started).
- Remove the onClick event and display tooltips instead of Toasts.

I have some demo videos here:
Before:

https://github.com/hyperlane-xyz/hyperlane-explorer/assets/15717476/67bc4d5a-c29f-4fb8-9a09-a96640b66da0

After:

https://github.com/hyperlane-xyz/hyperlane-explorer/assets/15717476/f515aa63-3e54-4d07-a9c6-aede39e808ab

### DIFF
--- a/src/components/icons/HelpIcon.tsx
+++ b/src/components/icons/HelpIcon.tsx
@@ -1,13 +1,12 @@
 import { memo } from 'react';
-import { toast } from 'react-toastify';
 
 import Question from '../../images/icons/question-circle.svg';
 import { IconButton } from '../buttons/IconButton';
 
 function _HelpIcon({ text, size = 20 }: { text: string; size?: number }) {
-  const onClick = () => {
-    toast.info(text, { autoClose: 8000 });
-  };
+  // const onClick = () => {
+  //   toast.info(text, { autoClose: 8000 });
+  // };
 
   return (
     <IconButton
@@ -15,8 +14,13 @@ function _HelpIcon({ text, size = 20 }: { text: string; size?: number }) {
       title="Help"
       width={size}
       height={size}
-      onClick={onClick}
+      // onClick={onClick}
       classes="opacity-50"
+      passThruProps={{
+        'data-tooltip-content': text,
+        'data-tooltip-id': 'my-tooltip',
+        'data-tooltip-place': 'top-start',
+      }}
     />
   );
 }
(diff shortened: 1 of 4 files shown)


src/pages/_app.tsx
```diff
@@ -84,8 +85,9 @@ export default function App({ Component, router, pageProps }: AppProps) {
         </UrqlProvider>
       </QueryClientProvider>
       <ToastContainer transition={Zoom} position={toast.POSITION.BOTTOM_RIGHT} limit={2} />
+      <Tooltip id="my-tooltip" className="z-50" />
```
jmrossy, line 88: Let's call this `"root-tooltip"` instead please

src/components/icons/HelpIcon.tsx
```diff
@@ -1,22 +1,26 @@
...
 import Question from '../../images/icons/question-circle.svg';
 import { IconButton } from '../buttons/IconButton';
 
 function _HelpIcon({ text, size = 20 }: { text: string; size?: number }) {
-  const onClick = () => {
-    toast.info(text, { autoClose: 8000 });
-  };
+  // const onClick = () => {
+  //   toast.info(text, { autoClose: 8000 });
+  // };
```
jmrossy, line 9: You can delete this code now

src/components/icons/HelpIcon.tsx
```diff
@@ -1,22 +1,26 @@
...
 
   return (
     <IconButton
       imgSrc={Question}
       title="Help"
       width={size}
       height={size}
-      onClick={onClick}
+      // onClick={onClick}
```
jmrossy, line 17: ditto

jmrossy (commented):
Looks great, thank you! Just a few minor tweaks needed

tiendn:
Thanks @jmrossy.
I've updated

jmrossy (approved):

PR #69, opened 2024-04-29
Feature/refactor try decode ica body function

NguyenTuanCanh:
Here are some suggestions for improvement:

1. Input Validation: Ensure that the input body is valid before attempting to decode it. It seems like you're checking if it's falsy or if it represents zero in a BigNumber context. However, it might be more robust to explicitly check if it's a string.
2. Error Handling: Your code is already catching errors during decoding, which is good practice. However, it's important to handle errors appropriately and provide meaningful error messages.
3. Type Safety: TypeScript provides powerful type checking capabilities. Ensure that you're leveraging them fully to catch potential errors at compile time.

### DIFF
--- a/src/features/messages/ica.ts
+++ b/src/features/messages/ica.ts
@@ -70,7 +70,7 @@ export function tryDecodeIcaBody(body: string): { sender: string; calls: { desti
 
     return { sender, calls: formattedCalls };
   } catch (error) {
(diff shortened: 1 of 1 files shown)


jmrossy:
Thanks @NguyenTuanCanh for the effort here but I don't see any real value in this change. Given that your change also fails to build and fails the formatter check, I don't think we should spend any more time on this. Sorry!

PR #71, opened 2024-04-30
fix: add time elapsed in transaction list (Home)

tiendn:
**Issues: UX performance about transaction list in page /home**

We should have a time elapsed in transaction list (Home Screen) for better UX.

Solutions: 
- Add a **TimeElapsed.tsx** component with setInterval each second for checking new value of readable time string
- Replace current function string to **TimeElapsed** component only on Home screen

https://github.com/hyperlane-xyz/hyperlane-explorer/assets/15717476/39d7fcdf-6cdb-40ce-85ce-acec3ee3b2e7

### DIFF
--- a/src/components/timer/TimeElapsed.tsx
+++ b/src/components/timer/TimeElapsed.tsx
@@ -1,0 +1,20 @@
+import { useEffect, useState } from 'react';
+
+import { getHumanReadableTimeString } from '../../utils/time';
+
+type TimeElapsedProps = {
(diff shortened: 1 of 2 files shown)


jmrossy:
Thanks for this contribution but it was intentional to keep the time sent static. We believe it's too much noise when all the numbers change constantly. But we'll consider your change and get back to you.

tiendn:
Okay, tks

PR #72, opened 2024-05-09
Migrate explorer to registry

jmrossy:
Fixes https://github.com/hyperlane-xyz/hyperlane-monorepo/issues/3655

- Migrate explorer from SDK consts to registry
- Update to latest hyperlane libs


package.json
```diff
@@ -1,12 +1,13 @@
...
+  "version": "3.11.0",
   "author": "J M Rossy",
   "dependencies": {
     "@headlessui/react": "^1.7.17",
-    "@hyperlane-xyz/sdk": "3.8.0",
-    "@hyperlane-xyz/utils": "3.8.0",
+    "@hyperlane-xyz/registry": "^1.1.0",
+    "@hyperlane-xyz/sdk": "3.11.1",
+    "@hyperlane-xyz/utils": "3.11.1",
     "@hyperlane-xyz/widgets": "3.8.0",
```
yorhodes, line 11: this doesnt use your widgets changes? seems like an opportunity to dedupe?

src/features/api/utils.ts
```diff
@@ -1,7 +1,16 @@
...
 export function successResult<R>(data: R): { success: true; data: R } {
   return { success: true, data };
 }
 
 export function failureResult(error: string): { success: false; error: string } {
   return { success: false, error };
 }
+
+export async function getMultiProvider(): Promise<MultiProvider> {
+  const registry = new GithubRegistry();
```
yorhodes, line 13: might be nice to consolidate this instance with the store instance so we can override the registry url in one place

yorhodes (approved):
nice

package.json
jmrossy: It will now! Just needed to merge and publish that part first

src/features/api/utils.ts
jmrossy: Agreed though there's some trickiness because stores are meant to be consumed via react hooks whereas this is server-side code. Leaving a TODO comment for now.

PR #90, opened 2024-06-12
Update SDK, registry, and widget libs to latest

jmrossy:
Related: https://discord.com/channels/935678348330434570/1008821009731829810/1248421471529730100


AlexBHarley (approved):

AlexBHarley (approved):
Approving with the caveat that I don't know how releases work for this. Are they manual or will Vercel pick this up automatically? 

Feel like for the Registry going V1 -> V2 it should probably be QA'd

PR #93, opened 2024-07-02
fix: error non-evm chain selected in filter

jmrossy:
Proxy for https://github.com/hyperlane-xyz/hyperlane-explorer/pull/92 to fix the prettier CI failure


src/components/search/SearchFilterBar.tsx
```diff
@@ -92,7 +92,8 @@ function ChainMultiSelector({
     );
     const mainnets = coreEvmChains.filter((c) => !c.isTestnet);
     const testnets = coreEvmChains.filter((c) => !!c.isTestnet);
-    return { chains, mainnets, testnets };
+    // Return only evnChains because of graphql only accept query non-evm chains (with bigint type not string)
```
paulbalaji, line 95:
```suggestion
    // Return only evmChains because of graphql only accept query non-evm chains (with bigint type not string)
```

paulbalaji (approved):

PR #98, opened 2024-07-16
fix: clear search state when route to Homepage with asPath = /

tiendn:
### Description
Currently, when a user performs a search and wants to reset or go back, they must manually clear the text or click the close button.
To improve convenience, clicking on the **Hyperlane Logo Header** should clear all input text.

### How
Add `useEffect` on **MessageSearch.tsx**, check when router changed asPath to `'/'`, set reset input in background

### DIFF
--- a/src/features/messages/MessageSearch.tsx
+++ b/src/features/messages/MessageSearch.tsx
@@ -1,4 +1,5 @@
-import { useState } from 'react';
+import { useRouter } from 'next/router';
+import { useEffect, useState } from 'react';
 
 import { Fade } from '../../components/animations/Fade';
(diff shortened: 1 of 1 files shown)


jmrossy:
@tiendn Thanks for the contribution but we feel that the UX benefit is minimal but this change complicates the URL query param management. Currently the logic is handled in `useSyncQueryParam` but this adds extra logic in the MessageSearch component.

tiendn:
Hm, OK.
I just think it makes the UX more convenient, but if there is extra logic, I’m not sure.
May I ask, what extra logic is there?

PR #99, opened 2024-07-24
chores: clean the code to make it more concise

dima-safin-2025:


src/utils/string.ts
```diff
@@ -4,16 +4,15 @@ import { strip0x } from '@hyperlane-xyz/utils';
 // Only allows letters and numbers
 const alphanumericRgex = /[^a-zA-Z0-9]/gi;
 export function sanitizeString(str: string) {
-  if (!str || typeof str !== 'string') return '';
```
jmrossy, line 7: Can please you restore the `typeof` check here?

jmrossy (commented):
Nice cleanup, thank you! Just one thing to fix before merging

src/utils/string.ts
dima-safin-2025: Of course, but I have a question . If  the parameter type of "str" is string in typescript , does typeof str !== 'string' always evaluate to false? 
32c1789e-9245-443f-ad4a-70d0db2d6368

jmrossy: Not necessarily. Typescript only does **build-time** checking. At **runtime**, it's still possible due to developer mistakes or invalid data somewhere for a non-string value to get passed into this function. That's why this defensive check is still useful :)

dima-safin-2025: I got it. I have already fixed this issue. Thanks for your patience in answering my questions.

jmrossy (approved):

jmrossy:
@dima-safin-2025 I've approved the changes but it's look like you'll need to run `yarn prettier --fix`. The code formatter has not been run on your changes.

dima-safin-2025:
> @dima-safin-2025 I've approved the changes but it's look like you'll need to run `yarn prettier --fix`. The code formatter has not been run on your changes.

I've run `yarn prettier --fix` on the affected files and committed the changes

PR #101, opened 2024-08-01
chore: update registry to v2.4.0

paulbalaji:
chore: update registry to v2.4.0


tkporter (approved):
have little context here but lgtm

PR #104, opened 2024-08-16
feat: Add information to msg transaction cards

jmrossy:
- Display duration in dest tx card per request from @Skunkchain 
- Display mailbox address in both tx cards
- Hide API docs page link which was disabled in #77 
- Fix missing type annotation


paulbalaji (commented):
generally lgtm, curious why the mailbox is unknown for this tx https://hyperlane-explorer-k2eikc3ff-abacus-works.vercel.app/message/0xc271e92077b44728043c61308f1e607cb6df4a21599ab52c25dccc1ac2095822

jmrossy:
@paulbalaji added protection for the case where the mailbox is not yet known

paulbalaji (approved):

PR #106, opened 2024-08-27
Get scraped chains list from DB instead of registry

jmrossy:
- Update Next and Hyperlane deps
- Query scraped chains from domains DB table

Fixes https://github.com/hyperlane-xyz/hyperlane-explorer/issues/74

Enabled us to potentially cut the CoreChains list from the registry (assuming the Infra package doesn't need it)

### DIFF
--- a/src/consts/config.ts
+++ b/src/consts/config.ts
@@ -1,5 +1,3 @@
-import { CoreChain } from '@hyperlane-xyz/registry';
-
 const isDevMode = process?.env?.NODE_ENV === 'development';
 const version = process?.env?.NEXT_PUBLIC_VERSION ?? null;
 const explorerApiKeys = JSON.parse(process?.env?.EXPLORER_API_KEYS || '{}');
@@ -17,7 +15,3 @@ export const config: Config = Object.freeze({
   apiUrl: 'https://explorer4.hasura.app/v1/graphql',
   explorerApiKeys,
 });
-
-// Based on https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/mainnet3/agent.ts
-// Based on https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/testnet4/agent.ts
-export const unscrapedEvmChains = [CoreChain.proteustestnet, CoreChain.sei, CoreChain.viction];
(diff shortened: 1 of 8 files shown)


src/consts/config.ts
```diff
@@ -17,7 +15,3 @@ export const config: Config = Object.freeze({
   apiUrl: 'https://explorer4.hasura.app/v1/graphql',
   explorerApiKeys,
 });
-
-// Based on https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/mainnet3/agent.ts
-// Based on https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/infra/config/environments/testnet4/agent.ts
-export const unscrapedEvmChains = [CoreChain.proteustestnet, CoreChain.sei, CoreChain.viction];
```
paulbalaji, line 23: there might still be an entry in the DB for unscraped chains, will that have a negative impact?

src/features/api/getMessages.ts
```diff
@@ -27,7 +27,9 @@ export async function handler(
   const result = await client.query<MessagesQueryResult>(query, variables).toPromise();
 
   const multiProvider = await getMultiProvider();
-  const messages = parseMessageQueryResult(multiProvider, result.data);
+  const scrapedChains = await getScrapedChains(client);
```
paulbalaji, line 30: unfamiliar with graphql, will this lead to multiple fetches of domains from the DB / how does the query's `@cached` help?

jmrossy: This isn't what the UI uses. This is the handler for the explorer's API, which is actually disabled right now anyway, but I'm trying to keep the code working. It will run once per API request as it should.

src/consts/config.ts
jmrossy: Ah maybe. I checked and proteustestnet is not in the domains table but sei and viction are. Why don't we scrape those? The impact would be that they should up in the chain list but show no messages when you filter to them. I can restore this list but it would be great if we can kill it eventually

paulbalaji: for sei, concerns about [pointer contracts](https://www.docs.sei.io/dev-interoperability/pointer-contracts) and how the scraper might not gracefully exit if it hits an issue there

for viction, their RPC non-compliance breaks scraping

will see about removing from the DB and once that's good we can come back and kill this

paulbalaji (approved):

PR #126, opened 2024-10-18
fix: header scroll looping and firefox slow animation

Xaroz:
## Scroll loop fix
Fixed an issue where on some specific threshold it would cause the header to loop the scaling animation because of the height changes. Move timeoutId so the variable is not re-created everytime and added a missing logic that would prevent proper debouncing

Caveat: If the scroll is extremely fast it might cause the header to be stuck on a certain threshold until a new scroll happens

## Firefox animation fix
Fixed an issue with `Firefox` choppy and slow animation, basically some browsers struggle with animation because of the way they render a page, so the "trick" is to make the browser go into hardware acceleration mode by adding a slight `rotation` and this way the animations will run smoother

On Firefox if you pay really close attention there's still a bit of a delay/blurriness that is almost not noticeable when the animation finishes

### DIFF
--- a/src/utils/useScrollListener.ts
+++ b/src/utils/useScrollListener.ts
@@ -1,47 +1,49 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 
-import { isFirefox } from './browser';
-
-export function useScrollThresholdListener(threshold: number, debounce = 500) {
+export function useScrollThresholdListener(threshold: number, debounceTime = 500) {
   const [isAboveThreshold, setIsAbove] = useState(false);
   const [isDebouncing, setIsDebouncing] = useState(false);
 
+  const handleScroll = useCallback(() => {
+    if (isDebouncing) return; // Skip handling scroll when disabled
+
+    if (window.scrollY > threshold && !isAboveThreshold) {
+      setIsAbove(true);
+      setIsDebouncing(true);
+    } else if (window.scrollY <= threshold && isAboveThreshold) {
+      setIsAbove(false);
+      setIsDebouncing(true);
+    }
+  }, [threshold, isAboveThreshold, isDebouncing]);
+
   useEffect(() => {
-    let timeoutId: NodeJS.Timeout | null;
+    const debouncedHandleScroll = debounce(handleScroll, 30);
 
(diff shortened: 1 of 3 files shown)


src/utils/useScrollListener.ts
```diff
@@ -1,47 +1,50 @@
...
+    if (window.scrollY > threshold && !isAboveThreshold) {
+      setIsAbove(true);
+      setIsDebouncing(true);
+    } else if (window.scrollY <= threshold && isAboveThreshold) {
+      setIsAbove(false);
+      setIsDebouncing(true);
+    }
+  }, [threshold, isAboveThreshold, isDebouncing]);
+
+  const debouncedHandleScroll = debounce(handleScroll, 20);
```
jmrossy, line 21: This will create a new fn on each render, making the `useCallback` around handleScroll's implementation moot, and causing the useEffect hook below to run on each render.

src/utils/useScrollListener.ts
```diff
@@ -1,47 +1,50 @@
...
     };
-  }, [threshold, debounce, isAboveThreshold, isDebouncing]);
+  }, [debouncedHandleScroll, isDebouncing, debounceTime]);
 
   return isAboveThreshold;
 }
+
+function debounce(fn: () => void, delay: number) {
+  let timeoutId: number;
+  return function () {
```
jmrossy, line 44: Use `function` to top-level fn definitions (for recursion support and better debugging) but arrow fns for nested ones (to have `this` bound properly)

src/utils/useScrollListener.ts
```diff
@@ -1,47 +1,50 @@
...
+    if (timeoutId) {
+      clearTimeout(timeoutId);
+    }
+    timeoutId = window.setTimeout(fn, delay);
```
jmrossy, line 48: Here you use `window.` but for other set/clear timeout calls you don't. Be consistent. You don't need `window.` here

jmrossy (changes requested):
Thanks for fixing this. It does seem to work but your `debounce` function is a bit sloppy and I don't think you're handling your callback memoization correctly yet. Please have a closer look.

jmrossy (approved):

PR #127, opened 2024-10-22
Update useScrollListener.ts

LyDe17TC:
I've removed the isDebouncing variable which was used to track whether the delay was active. Instead, timeout management is now controlled solely by timeoutId. This simplifies the code, as we no longer need to check the status of the offset each time. The timer (timeoutId) takes care of managing the wait between checks.

### DIFF
--- a/src/utils/useScrollListener.ts
+++ b/src/utils/useScrollListener.ts
@@ -2,34 +2,24 @@ import { useEffect, useRef, useState } from 'react';
 
 export function useScrollThresholdListener(threshold: number, debounce = 300) {
   const [isAboveThreshold, setIsAbove] = useState(false);
-  const [isDebouncing, setIsDebouncing] = useState(false);
 
   const timeoutId = useRef<NodeJS.Timeout | null>(null);
 
   useEffect(() => {
     const listener = () => {
       const handleScroll = () => {
-        if (isDebouncing) return;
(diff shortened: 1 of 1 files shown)


src/utils/useScrollListener.ts
```diff
@@ -2,34 +2,24 @@ import { useEffect, useRef, useState } from 'react';
 
 export function useScrollThresholdListener(threshold: number, debounce = 300) {
   const [isAboveThreshold, setIsAbove] = useState(false);
-  const [isDebouncing, setIsDebouncing] = useState(false);
```
Xaroz, line 5: Not sure entirely about removing the debouncing variable, by doing so it causes certain threshold to just call handleScroll in a loop

Xaroz (changes requested):

src/utils/useScrollListener.ts
jmrossy: Yeah, as @Xaroz said, this debounce checking was to avoid an infinite loop where the header animation would change size, triggering the listener, which would change the size, etc.

Thanks @LyDe17TC for the contribution but we'll leave this hook as it is.

LyDe17TC: no problem, if I can help you don't hesitate

PR #130, opened 2024-10-23
fix: header stuck on certain threshold

Xaroz:
On some cases if the scroll happens fast and multiples times, it will get stuck on a threshold and not call the `handleScroll` again, by cleaning the timeoutId everytime `handleScroll` gets called we ensure that it will not get stuck if scrolling happens too fast.


jmrossy (approved):
Thanks!

PR #136, opened 2024-11-06
feat: use chain name or domain as identifier instead of chainId 

paulbalaji:
relates to:
- https://github.com/hyperlane-xyz/hyperlane-warp-ui-template/pull/313
- https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/4798

testing:
- [x] update SDK to latest once monorepo PR is merged + rolled out
- [x] add duplicate chain metadata with different domain id to custom yaml

### DIFF
--- a/src/components/search/SearchFilterBar.tsx
+++ b/src/components/search/SearchFilterBar.tsx
@@ -70,8 +70,11 @@ function ChainSelector({
   const { isOpen, open, close } = useModal();
 
   const multiProvider = useMultiProvider();
-  const chainName = value
-    ? trimToLength(getChainDisplayName(multiProvider, value, true), 12)
+
+  // TODO: chain selector should use domainId
+  const chainRawName = value ? multiProvider.getChainName(value) : undefined;
+  const chainName = chainRawName
+    ? trimToLength(getChainDisplayName(multiProvider, chainRawName, true), 12)
     : undefined;
 
   const onClickChain = (c: ChainMetadata) => {
--- a/src/components/search/SearchFilterBar.tsx
+++ b/src/components/search/SearchFilterBar.tsx
@@ -71,9 +71,9 @@ function ChainSelector({
 
   const multiProvider = useMultiProvider();
 
-  const chainRawName = value ? multiProvider.getChainName(value) : undefined;
-  const chainName = chainRawName
-    ? trimToLength(getChainDisplayName(multiProvider, chainRawName, true), 12)
+  const chainName = value ? multiProvider.getChainName(value) : undefined;
+  const chainDisplayName = chainName
+    ? trimToLength(getChainDisplayName(multiProvider, chainName, true), 12)
     : undefined;
 
   const onClickChain = (c: ChainMetadata) => {
@@ -95,7 +95,7 @@ function ChainSelector({
         )}
         onClick={open}
       >
-        <span>{chainName || text} </span>
+        <span>{chainDisplayName || text} </span>
         {!value && (
           <ChevronIcon
             direction="s"
--- a/src/components/search/SearchFilterBar.tsx
+++ b/src/components/search/SearchFilterBar.tsx
@@ -64,14 +64,13 @@ function ChainSelector({
   onChangeValue,
 }: {
   text: string;
-  value: ChainId | null;
+  value: string | null;
   onChangeValue: (value: string | null) => void;
 }) {
   const { isOpen, open, close } = useModal();
 
   const multiProvider = useMultiProvider();
 
-  // TODO: chain selector should use domainId
   const chainRawName = value ? multiProvider.getChainName(value) : undefined;
   const chainName = chainRawName
     ? trimToLength(getChainDisplayName(multiProvider, chainRawName, true), 12)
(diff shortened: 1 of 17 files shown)


src/components/search/SearchFilterBar.tsx
```diff
@@ -64,8 +64,11 @@ function ChainSelector({
   const { isOpen, open, close } = useModal();
 
   const multiProvider = useMultiProvider();
-  const chainName = value
-    ? trimToLength(getChainDisplayName(multiProvider, value, true), 12)
+
+  // TODO: chain selector should use domainId
```
paulbalaji, line 68: not immediately clear how to update, will look into it

@jmrossy any pointers here?

src/features/chains/queries/useScrapedChains.ts
```diff
@@ -42,8 +42,8 @@ export function useScrapedChains(multiProvider: MultiProvider) {
     const scrapedChains = objFilter(
       chainMetadata,
       (_, chainMetadata): chainMetadata is ChainMetadata =>
-        !isPiChain(multiProvider, scrapedDomains, chainMetadata.chainId) &&
-        !isUnscrapedDbChain(multiProvider, chainMetadata.chainId),
+        !isPiChain(multiProvider, scrapedDomains, chainMetadata.domainId!) &&
```
paulbalaji, line 45: the `domainId!` will be simplified to just `domainId` with an SDK update

src/features/chains/queries/useScrapedChains.ts
```diff
@@ -52,7 +52,7 @@ export function useScrapedChains(multiProvider: MultiProvider) {
 }
 
 // TODO: Remove once all chains in the DB are scraped
-export function isUnscrapedDbChain(multiProvider: MultiProvider, chainIdOrName: number | string) {
-  const chainName = multiProvider.tryGetChainName(chainIdOrName);
+export function isUnscrapedDbChain(multiProvider: MultiProvider, domainId: DomainId) {
```
paulbalaji, line 55: had to change a bunch of `chainIdOrName` because it's now just `domainId` or `chainName` in places

src/components/search/SearchFilterBar.tsx
jmrossy: It already does use domainId doesn't it? See the onCLickChain handler just below

src/features/messages/pi-queries/fetchPiChainMessages.ts
```diff
@@ -127,23 +127,23 @@ async function fetchLogsForAddress(
...
 async function fetchLogsForTxHash(
-  { chainId }: ChainMetadata,
+  { domainId }: ChainMetadata,
   query: PiMessageQuery,
   multiProvider: MultiProvider,
 ): Promise<ExtendedLog[]> {
   const txHash = query.input;
-  logger.debug(`Fetching logs for txHash ${txHash} on chain ${chainId}`);
-  const provider = multiProvider.getProvider(chainId);
+  logger.debug(`Fetching logs for txHash ${txHash} on chain ${domainId}`);
```
jmrossy, line 142: since we have the `ChainMetadata` object here anyway, I wonder if we should log the chain name instead. Would be more helpful for debugging. I would have done that in the first place but, like the warp ui, the first version of this app used chainId everywhere because that's what the broader standard is

src/utils/explorers.ts
```diff
@@ -152,7 +152,7 @@ export async function queryExplorerForTx(
 
 export async function queryExplorerForTxReceipt(
   multiProvider: MultiProvider,
-  chainId: ChainId,
+  domainId: DomainId,
```
jmrossy, line 155: chainName would probably be better if we're going through the trouble of changing these but def ndb

jmrossy (commented):
Looking good so far!

src/features/messages/pi-queries/fetchPiChainMessages.ts
paulbalaji: I think `name (domain)` would work well

src/utils/explorers.ts
paulbalaji: i tried to minimise diff but yeah happy to go with chainName - it's my preferred key too

src/components/search/SearchFilterBar.tsx
paulbalaji: clarified - just misleading type

src/components/search/SearchFilterBar.tsx
```diff
@@ -58,14 +58,16 @@ function ChainSelector({
...
+  value: string | null;
   onChangeValue: (value: string | null) => void;
 }) {
   const { isOpen, open, close } = useModal();
 
   const multiProvider = useMultiProvider();
-  const chainName = value
-    ? trimToLength(getChainDisplayName(multiProvider, value, true), 12)
+
+  const chainRawName = value ? multiProvider.getChainName(value) : undefined;
```
jmrossy, line 68: please call this `chainName` and rename the next var to `chainDisplayName`

jmrossy (approved):

PR #147, opened 2024-12-10
feat: Style improvements for buttons and tooltips

NikolaiKryshnev:
[Fix: Ensure proper tooltip rendering on mobile devices](https://github.com/hyperlane-xyz/hyperlane-explorer/commit/1a96011df0cdc2caf66cac61d886cedcd0d53b13)

BEFORE: 

https://github.com/user-attachments/assets/7eb54f21-5ff4-4ef6-ba46-62132ca3220c

AFTER

https://github.com/user-attachments/assets/291e2fd4-3c29-4473-a27a-7af1db77e6ea

[Enhance button visibility with a hover effect (pink to blue transition).](https://github.com/hyperlane-xyz/hyperlane-explorer/commit/f52207d9bd5636816ff5c373fa8e3e60f27ffb14)

BEFORE: 

https://github.com/user-attachments/assets/dc7805d3-2b3a-4d22-966e-864acf975d5f

AFTER: 

https://github.com/user-attachments/assets/c5ae04b5-ea33-4ec1-b27b-0a62433ed38e

[Fixed: Button now covers parent area for intuitive usability and incl…](https://github.com/hyperlane-xyz/hyperlane-explorer/commit/36136a54444236b32b60a97d5ae323e17152ddd0)

BEFORE: 

https://github.com/user-attachments/assets/2b03e7f3-4d36-400e-b503-355e71425b65

AFTER:

https://github.com/user-attachments/assets/94fa3d77-049a-41a1-86b1-a2ff03154eb9

### DIFF
--- a/src/components/search/SearchBar.tsx
+++ b/src/components/search/SearchBar.tsx
@@ -19,6 +19,9 @@ export function SearchBar({ value, placeholder, onChangeValue, isFetching }: Pro
     onChangeValue(value);
   };
 
+  const iconStyle =
+    'flex h-10 w-10 items-center justify-center rounded-full bg-pink-600 sm:h-12 sm:w-12';
+
   return (
     <div className="flex w-full items-center rounded-full bg-white p-1 transition-all duration-500">
       <input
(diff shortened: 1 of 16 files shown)


src/components/search/SearchBar.tsx
```diff
@@ -19,6 +19,9 @@ export function SearchBar({ value, placeholder, onChangeValue, isFetching }: Pro
     onChangeValue(value);
   };
 
+  const iconStyle =
```
jmrossy, line 22: Please move this to the bottom of the file into an object called `styles`. Here's an example: https://github.com/hyperlane-xyz/hyperlane-explorer/blob/main/src/components/nav/Footer.tsx#L93

src/styles/custom.css
```diff
@@ -0,0 +1,4 @@
+/* Fix for tooltips breaking layout on mobile devices (<526px width). */
+
+
+@media (max-width: 526px) {#gas-info, #transaction-info, #message-info{max-width: calc(100% - 10px);}}
```
jmrossy, line 4: Please use tailwind classes instead of creating a new css file.
If you need to do specific amounts like 526px, you can use a class like `max-w-[526px]`, but it would be better if you can make this work with just the `max-w` classes in tailwind's base set.

jmrossy (commented):
Thanks for these improvements! Just a few small things to fix please

src/components/search/SearchBar.tsx
NikolaiKryshnev: Done!)

src/styles/custom.css
NikolaiKryshnev: Thanks for the suggestion! Unfortunately, this appears to be a limitation or bug in the Tooltip library. When we try to pass a className like max-w-[calc(100%-10px)] sm:max-w-[526px], it gets applied to an unexpected <a> tag instead of the actual tooltip. This behavior prevents us from fully leveraging Tailwind classes in this case. For now, me had to rely on a custom CSS file to address this issue.

Xaroz: There is a property inside the Tooltip called `data-tooltip-class-name` which directly modifies the `className` for the tooltip component. I vote we also move this part to the widgets @jmrossy

jmrossy: @NikolaiKryshnev would you like to give this classname-based approach a try? Also from the CI failure it looks like you need to run `yarn prettier` to fix the formatting issue

jmrossy (approved):
Looks like prettier is still failing due to the extra space is SearchBar but I'll merge as is and fix it for you.

PR #149, opened 2024-12-12
fix: disabled bytecode handler and add ignored chains for debugger

Xaroz:
- Comment out bytecode handler for less specific error messages
- Added debugIgnoredChains for skipping some specifics chains
- New fallback for skipping some chains in `debugMessage`, will show spinner until messages actually delivers or fails

image

### DIFF
--- a/src/features/debugger/debugMessage.ts
+++ b/src/features/debugger/debugMessage.ts
@@ -186,8 +186,8 @@ async function debugMessageDelivery(
 
     if (debugIgnoredChains.includes(destName)) {
       return {
-        status: MessageDebugStatus.MessageNotDelivered,
-        description: 'Message not delivered, there may be an error',
+        status: null,
+        description: '',
         calldataDetails,
       };
     }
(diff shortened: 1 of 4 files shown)


src/features/debugger/debugMessage.ts
```diff
@@ -172,12 +174,20 @@ async function debugMessageDelivery(
...
+    //     calldataDetails,
+    //   };
+    // }
+
+    if (debugIgnoredChains.includes(destName)) {
       return {
-        status: MessageDebugStatus.RecipientNotHandler,
-        description: `Recipient contract should have handle function of signature: ${HANDLE_FUNCTION_SIG}. Check that recipient is not a proxy. Error: ${errorReason}`,
+        status: MessageDebugStatus.MessageNotDelivered,
+        description: 'Message not delivered, there may be an error',
```
jmrossy, line 190: Rather than create this new case that doesn't really communicate anything, I'd prefer if you just never called  `debugMessage` for the blacklisted chains please

jmrossy (approved):

PR #153, opened 2024-12-16
fix: status icon not showing on mobile

Xaroz:
Fixed an issue that caused the icons to not show for mobile screens

- The culprit is the `absolute` position on the icon span, that anchors to the main table `relative` (Kinda weird that it worked before). All the icons would absolute position to the center of the table
- Removed `absolute` position on the icon, added a small width to the `LinkCell` so that it can display the icon

IMG_6387

### DIFF
--- a/src/features/messages/MessageTable.tsx
+++ b/src/features/messages/MessageTable.tsx
@@ -96,12 +96,12 @@ export function MessageSummaryRow({ message, mp }: { message: MessageStub; mp: M
       >
         {shortenAddress(origin.hash)}
       </LinkCell>
-      <LinkCell id={msgId} base64={base64} aClasses={styles.valueTruncated} tdClasses="pr-5">
+      <LinkCell id={msgId} base64={base64} aClasses={styles.valueTruncated}>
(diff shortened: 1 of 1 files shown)


src/features/messages/MessageTable.tsx
```diff
@@ -99,9 +99,9 @@ export function MessageSummaryRow({ message, mp }: { message: MessageStub; mp: M
       <LinkCell id={msgId} base64={base64} aClasses={styles.valueTruncated} tdClasses="pr-5">
         {getHumanReadableTimeString(origin.timestamp)}
       </LinkCell>
-      {statusIcon && (
-        <LinkCell id={msgId} base64={base64} tdClasses="w-0">
-          <span className="absolute right-3 top-1/2 -translate-y-1/2 transform">
+      <LinkCell id={msgId} base64={base64} tdClasses="w-8">
+        {statusIcon && (
+          <span className="-translate-y-1/2 transform">
```
Xaroz, line 104: Moved this like this because on previous iteration, since we are not using absolute positioning anymore, it would cause borders to be missing.

jmrossy (approved):

PR #154, opened 2024-12-20
Update explorers.ts

nikbenesh:
Added some error handling, type safety and documentation.  Since there’s a TODO about deduplication with another library, consider creating a shared utility module that both files can import from once the widgets library exports those utilities.

### DIFF
--- a/src/utils/explorers.ts
+++ b/src/utils/explorers.ts
@@ -1,18 +1,14 @@
 // TODO de-dupe this file with widgets lib's utils/explorers.ts
 // The widgets lib doesn't export those yet, need to fix that first.
 import { BigNumber, providers } from 'ethers';
-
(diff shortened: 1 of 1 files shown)


src/utils/explorers.ts
```diff
@@ -26,43 +22,42 @@ async function queryExplorer<P>(
   chainName: string,
   params: URLSearchParams,
   useKey = false,
-) {
+): Promise<P> {
   const baseUrl = multiProvider.tryGetExplorerApiUrl(chainName);
-  if (!baseUrl) throw new Error(`No valid URL found for explorer for chain ${chainName}`);
+  if (!baseUrl) throw new Error(No valid URL found for explorer for chain ${chainName});
```
paulbalaji, line 27: this change makes no sense

PR #168, opened 2025-01-24
feat: refresh list

### LINKED ISSUE #157: Refresh button on explorer list page

Xaroz:
Add refresh button to message listing page, this will trigger a refetch for the graphQL query
image

fixes #157


src/features/messages/MessageSearch.tsx
```diff
@@ -138,16 +139,29 @@ export function MessageSearch() {
...
+              onClick={refetch}
+              className="h-7 w-7 rounded-lg bg-pink-500"
+              disabled={isAnyFetching}
+            >
+              {isAnyFetching ? (
+                <SpinnerIcon color="white" height={16} width={16} />
+              ) : (
+                <RefreshIcon color="white" height={20} width={20} />
+              )}
+            </IconButton>
```
jmrossy, line 163: Nit: I'd slightly prefer this be in it's own component. It can be just lower in this file

src/features/messages/MessageSearch.tsx
```diff
@@ -138,16 +139,29 @@ export function MessageSearch() {
...
+              destinationChain={destinationChainFilter}
+              onChangeDestination={setDestinationChainFilter}
+              startTimestamp={startTimeFilter}
+              onChangeStartTimestamp={setStartTimeFilter}
+              endTimestamp={endTimeFilter}
+              onChangeEndTimestamp={setEndTimeFilter}
+            />
+            <IconButton
+              onClick={refetch}
+              className="h-7 w-7 rounded-lg bg-pink-500"
```
jmrossy, line 155: Nit: Do we need to explicitly set h and w here or can we use padding instead?

jmrossy (commented):
mostly looks good. Lmk when you figure out the yarn.lock issue. I see the change is with the TS package. Maybe it would help to pin the ts lib to a specific version

src/features/messages/MessageSearch.tsx
Xaroz: used padding instead

src/features/messages/MessageSearch.tsx
```diff
@@ -168,3 +172,15 @@ export function MessageSearch() {
     </>
   );
 }
+
+function RefreshButton({ loading, onClick }: { loading: boolean; onClick: () => void }) {
+  return (
+    <IconButton onClick={onClick} className="rounded-lg bg-pink-500 p-1" disabled={loading}>
+      {loading ? (
+        <SpinnerIcon color="white" height={16} width={16} />
```
jmrossy, line 180: I tested this and though the styles look fine, I do find this spinner distracting because it's right below the bigger spinner in the search bar and they run at the same time. WDYT? My preference would be to simply remove it.

jmrossy (approved):

src/features/messages/MessageSearch.tsx
Xaroz: yeah makes sense, I'll make adjustments and then keep the disabled part only

PR #144, opened 2024-11-27
Fix #78: Decode warp route messages.

spilehchiha:
This commit is a good base case for fixing https://github.com/hyperlane-xyz/hyperlane-explorer/issues/78.

Three edge cases remain to improve the UX:

1. In the case of transferring ERC-721 tokens
2. In the case of no value transfer transactions
3. Correctly showing the token symbol on the UX

Will work on this.

PR #176, opened 2025-02-04
feat: decode warp route information

### LINKED ISSUE #78: Decode warp route messages
## Problem

- Warp route messages are opaque on the explorer

## Solution

- Leverage https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/3784 to decode message bodies for warp route messages

### User story:
For users in the Explorer, see warp route data: which token, the amount of it, and the fees paid.

## Nice to Have

- [ ] Fetch warp route from registry and render additional metadata (eg collateral token metadata, chains connected, etc?)

Xaroz:
Leverage #144, to implement this:

Decodes warp route message to show more information about the warp token transfer
Fixes #78 

- Update the store to include a WarpRouteChainAddressMap, this is a custom type that is of `ChainMap<Record<Address, TokenArgs>>;`, the last string would be the token symbol
- Create `parseWarpRouteDetails()` function to decode the information about the warp transfer, a few checks are done to identify if the message is an actual warp route transfer
- Upgrade `hyperlane` packages

image
image


src/features/messages/queries/parse.ts
```diff
@@ -37,6 +44,70 @@ export function parseMessageQueryResult(
...
+    const chain = warpRouteChainAddressMap[name];
+    if (objKeys(chain).includes(address)) {
+      return chain[address];
+    }
+  }
+
+  return undefined;
+}
+
+export function parseWarpRouteDetails(
```
jmrossy, line 63: It maybe made sense to include this here when it was being called by `queryResult` but if it's only used by `WarpTransferDetailCard` then I'd argue it doesn't belong here. The code in this file takes DB data and turns it into the explorer's models. This fn takes the explorer's model and return warp route details. Doesn't feel cohesive with the rest of the code here. WDYT?

src/types.ts
```diff
@@ -43,6 +44,8 @@ export interface MessageStub {
   origin: MessageTxStub;
   destination?: MessageTxStub;
   isPiMsg?: boolean;
+  originMetadata?: ChainMetadata | null;
+  destinationMetadata?: ChainMetadata | null;
```
jmrossy, line 48: See comment above, recommend removing

src/features/messages/cards/WarpTransferDetailCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+            id="warp-route-info"
+            content="Information about the warp route transfer such as the end recipient and amount transferred"
+          />
+        </div>
+      </div>
+      <div className="flex flex-wrap gap-x-6 gap-y-4">
+        <KeyValueRow
+          label="Origin token:"
+          labelWidth="w-20 sm:w-32"
+          display={warpRouteDetails.originTokenAddress}
```
jmrossy, line 42: Nit: rather than repeat the `warpRouteDetails.` name in each KVRow here, destructure it above.

src/features/messages/queries/parse.ts
```diff
@@ -37,6 +44,70 @@ export function parseMessageQueryResult(
...
+      destinationMetadata,
+      recipient,
+      warpRouteChainAddressMap,
+    );
+
+    // If token symbols are not found with the addresses, it means the message is not a warp route transfer
+    if (!originTokenSymbol || !destinationTokenSymbol) return undefined;
+
+    const parsedMessage = parseWarpRouteMessage(body);
+    const address = postgresByteaToAddress(parsedMessage.recipient, destinationMetadata);
```
jmrossy, line 94: Why are you using `postgresByteaToAddress` here? That's for converting data from a postgres db to an address. The `parsedMessage` data is coming from the utility in `utils` right? Why would it be encoded like in the postgres db?

src/features/messages/queries/parse.ts
```diff
@@ -37,6 +44,70 @@ export function parseMessageQueryResult(
...
+
+    return {
+      amount: fromWei(parsedMessage.amount.toString(), originMetadata.nativeToken?.decimals || 18),
```
jmrossy, line 97: I think this is incorrect. The decimals for the warp token may not match the decimals for the native token. E.g. native ETH uses 18 but USDC uses 6.
You need the warp token's decimals here, which should be in the warp config

src/features/messages/queries/parse.ts
```diff
@@ -37,6 +44,70 @@ export function parseMessageQueryResult(
...
+      totalPayment: fromWei(totalPayment, originMetadata.nativeToken?.decimals || 18),
```
jmrossy, line 98: I think you're maybe misunderstanding what totalPayment refers to. It's a properly of the message relating to IGP gas. It's not related to warp transfer amounts

src/features/messages/queries/parse.ts
```diff
@@ -37,6 +44,70 @@ export function parseMessageQueryResult(
...
+      endRecipient: address,
```
jmrossy, line 99:
```suggestion
      recipient: address,
```

jmrossy (changes requested):
Good progress but I suspect there are some issues still to fix

src/features/messages/cards/WarpTransferDetailCard.tsx
```diff
@@ -0,0 +1,95 @@
+import Image from 'next/image';
+
```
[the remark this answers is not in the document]
Xaroz: done, also did this for files where I saw there are still spaces

src/features/messages/queries/parse.ts
```diff
@@ -90,6 +161,8 @@ function parseMessageStub(
           }
         : undefined,
       isPiMsg,
+      originMetadata,
+      destinationMetadata,
```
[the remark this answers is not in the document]
Xaroz: removed and used multiProvider to get these

src/store.ts
```diff
@@ -118,3 +128,22 @@ async function buildMultiProvider(
...
+      tokens.forEach(({ chainName, addressOrDenom, symbol }) => {
+        if (!acc[chainName]) {
+          acc[chainName] = {};
+        }
+        acc[chainName][addressOrDenom] = symbol;
+      });
+      return acc;
+    }, {});
+  } catch {
+    return {};
```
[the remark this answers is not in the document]
Xaroz: Remove try catch for now because we are not doing it `async` for now

src/store.ts
```diff
@@ -118,3 +128,22 @@ async function buildMultiProvider(
...
+  registry: IRegistry,
+): Promise<WarpRouteChainAddressMap> {
+  try {
+    const warpRoutes = await registry.getWarpRoutes();
+    return Object.values(warpRoutes).reduce((acc, { tokens }) => {
+      tokens.forEach(({ chainName, addressOrDenom, symbol }) => {
+        if (!acc[chainName]) {
+          acc[chainName] = {};
+        }
+        acc[chainName][addressOrDenom] = symbol;
```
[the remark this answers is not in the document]
Xaroz: Changed symbol to use token

src/features/messages/queries/parse.ts
Xaroz: Separated functions so that it makes more sense

src/features/messages/queries/parse.ts
Xaroz: Removed totalPayment and let `GasDetailCards` handle this

src/features/messages/queries/parse.ts
Xaroz: Renamed to transfer recipient

jmrossy (approved):

PR #180, opened 2025-02-17
fix: error in the Tailwind CSS class 

EricNorberg:
The "all" prefix is not a valid modifier in Tailwind CSS


Xaroz (changes requested):
Thanks for the PR! But the changes in this PR are not really significant so I am closing this

PR #184, opened 2025-02-26
fix: simplify null/undefined check for readability

mdqst:
I’ve updated the error-checking condition to make it more concise and readable. The previous check was checking explicitly for both `null` and `undefined`, but in JavaScript, `result == null` already covers both cases. By switching to `if (result == null)`, we reduce redundancy while maintaining the same behavior.

### DIFF
--- a/src/utils/number.ts
+++ b/src/utils/number.ts
@@ -13,6 +13,6 @@ export function tryToDecimalNumber(value: BigNumberish) {
 
 export function toDecimalNumber(value: BigNumberish) {
   const result = tryToDecimalNumber(value);
-  if (result === null || result === undefined) throw new Error(`Error parsing hex number ${value}`);
+  if (result == null) throw new Error(`Error parsing number ${value}`);
   return result;
 }


src/utils/number.ts
```diff
@@ -13,6 +13,6 @@ export function tryToDecimalNumber(value: BigNumberish) {
 
 export function toDecimalNumber(value: BigNumberish) {
   const result = tryToDecimalNumber(value);
-  if (result === null || result === undefined) throw new Error(`Error parsing hex number ${value}`);
+  if (result == null) throw new Error(`Error parsing number ${value}`);
```
Xaroz, line 16: javascript uses `===` for equality check and the reason this is checking for both null and undefined it is because we take into account `0` which is falsy value

Xaroz (changes requested):

Xaroz:
Thanks for the PR! While this is a good change, for some people it might be more confusing to read/understand, rather keep it the simple way

Issue #65, opened 2024-03-14
Smoother incremental additions to feed


jmrossy:
Instead of the simple blur and replace we have now, instead have the new messages add in and smoothly push older messages down

butcher73:
I've submitted PR #186 to implement the smooth message push-down effect. Initially, the table movement was a bit too abrupt, so I added a subtle animation to make it more fluid.

PR #186, opened 2025-03-03
Smoother incremental additions to feed #65

butcher73:
Highlight new messages, add loading row, and apply opacity during fetch. Improves message table UX with clear visual cues for updates and loading states.


Xaroz (commented):
Thank for the PR! Mostly looks good, only nit would be that I feel that when the page first loads all the message sliding down looks a bit abrupt and if you navigate to a message detail and then go back, the animation triggers again making it seem like there are new messages

PR #189, opened 2025-03-13
feat: link for block explorer addresses

Xaroz:
This PR adds a link for some of addresses that are in the Message Details and Warp Transfer Details section

- Adds link to block explorer for certain addresses
- Add LinkIcon to `KeyValueRow` component
- Add utils function `tryGetBlockExplorerAddressUrl` to retrieve block explorer address URL from `multiprovider`

### DIFF
--- a/src/utils/url.ts
+++ b/src/utils/url.ts
@@ -1,0 +1,12 @@
+import { ChainNameOrId, MultiProvider } from '@hyperlane-xyz/sdk';
+import BigNumber from 'bignumber.js';
+
+export async function tryGetBlockExplorerAddressUrl(
+  multiProvider: MultiProvider,
+  chain: ChainNameOrId,
+  address: string,
+) {
+  return address && !new BigNumber(address).isZero()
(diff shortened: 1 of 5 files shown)


src/utils/url.ts
```diff
@@ -0,0 +1,12 @@
+import { ChainNameOrId, MultiProvider } from '@hyperlane-xyz/sdk';
+import BigNumber from 'bignumber.js';
+
+export async function tryGetBlockExplorerAddressUrl(
+  multiProvider: MultiProvider,
+  chain: ChainNameOrId,
+  address: string,
+) {
+  return address && !new BigNumber(address).isZero()
```
paulbalaji, line 9: consider this helper from the utils package https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/583ea2525fde7843a66a6fe6069017ee737c4a15/typescript/utils/src/addresses.ts#L220

cmcewen (approved):
lgtm

PR #194, opened 2025-04-04
chore: fix quotation marks

taozui472:
The unexpected marks might confuse readers and be distracted by them. Also, some editors might not render correctly depending on the software or platform used, appearing as gibberish or causing formatting problems.


Xaroz:
Thanks for the PR but the LICENSE file is generated and this change is not currently needed

PR #196, opened 2025-04-14
feat: filter out testnet messages by default

### LINKED ISSUE #195: Exclude testnet from default messages
exclude testnet messages from home screen and only show when filters are applied

Xaroz:
close #195 

- Remove testnet message by default
- Now origin and destination will filter by mainnet chains if no filters are provided
- Bump `package.json` version in accordance to `hyperlane` packages upgrade

Running the queries locally show that they still get result very fast regardless of the amount of chains set (around 130 for both origin and destination)

Speed seems to run between: 90 ~ 360ms, where the highest spike occurs when the query is first fetched or if there hasn't been a refetch in a while, in comparison to the current production build where query last around 90~200ms

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@hyperlane-xyz/explorer",
   "description": "An interchain explorer for the Hyperlane protocol and network.",
-  "version": "7.1.0",
+  "version": "12.0.0",
   "author": "J M Rossy",
   "dependencies": {
     "@headlessui/react": "^2.2.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@hyperlane-xyz/explorer",
   "description": "An interchain explorer for the Hyperlane protocol and network.",
-  "version": "7.1.0",
+  "version": "12.0.0",
```
paulbalaji, line 4: why does this get bumped

paulbalaji (approved):

package.json
Xaroz: we forgot to bump this, but it seems that Rossy would bump the app version whenever `hyperlane` packages were upgraded

paulbalaji: coolio sgtm

PR #202, opened 2025-05-02
chore: safe parse chain metadata buildMultiprovider

Xaroz:
Now safeParse each chain before building multiProvider, this will prevent crashes related to the chain metadata

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -1,5 +1,11 @@
 import { GithubRegistry, IRegistry, warpRouteConfigs } from '@hyperlane-xyz/registry';
-import { ChainMap, ChainMetadata, MultiProvider, mergeChainMetadataMap } from '@hyperlane-xyz/sdk';
+import {
+  ChainMap,
+  ChainMetadata,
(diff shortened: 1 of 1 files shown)


src/store.ts
```diff
@@ -121,8 +127,12 @@ async function buildMultiProvider(
...
     ),
   );
-  const mergedMetadata = mergeChainMetadataMap(metadataWithLogos, overrideChainMetadata);
-  return { metadata: metadataWithLogos, multiProvider: new MultiProvider(mergedMetadata) };
+
+  const mergedMetadata = objFilter(
+    mergeChainMetadataMap(metadataWithLogos, overrideChainMetadata),
+    (_, metadata): metadata is ChainMetadata => ChainMetadataSchema.safeParse(metadata).success,
+  );
+  return { metadata: mergedMetadata, multiProvider: new MultiProvider(mergedMetadata) };
```
paulbalaji, line 135: we should probably log the error somewhere as well

paulbalaji (approved):

PR #208, opened 2025-05-13
chore: improve styling for warped route token

Xaroz:
- Fix img size issue for token icon
- Improve warped token row positioning
- Remove tooltip for token row and use empty space instead
- Now will show token icon for Warp Transfer Detail image if logoURI it exists
Screenshot 2025-05-13 at 10 17 48 AM
image
Screenshot 2025-05-13 at 10 17 54 AM


paulbalaji (approved):
style lgtm

PR #253, opened 2026-01-15
feat: add real-time validator signature status display

nambrot:
## Summary
- add `/api/ism-details` for server-side ISM + hook config derivation with best-effort caching
- vendor a small local metadata builder/type slice for multisig, aggregation, routing, and null ISMs
- surface validator signature status in the ISM details card, timeline, and pending destination tx path
- include the small review-followup cleanup for helper extraction and message-stage derivation

## Why
The debugger currently makes it hard to tell whether a destination message is still waiting on validator signatures, whether quorum is close, and which validators have already signed. This adds a server-derived ISM details path and renders that state directly in the message UI.

I intentionally removed the direct `@hyperlane-xyz/relayer` dependency from this PR and mirrored the small read-only metadata slice locally instead. `#253` only needs metadata building for validator-status display, but pulling in the published relayer package causes a large lockfile diff today because it drags in service deps and satisfies optional logging peers on the current Hyperlane package line. Keeping the metadata slice local keeps this PR reviewable and avoids mixing metadata UI work with relayer package churn. The full relayer dependency can land in `#247`, where it is actually needed for self-relay.

## Stack
- base PR: `#301` (`perf: keep explorer metadata-first and evm-only at runtime`)
- this PR should merge after `#301`

## Validation
- `pnpm run format`
- `pnpm run typecheck`
- `pnpm run build`


src/features/messages/cards/IsmDetailsCard.tsx
```diff
@@ -32,32 +40,292 @@ export function IsmDetailsCard({ ismDetails, blur }: Props) {
...
+          {/* Signed progress */}
+          {signedCount > 0 && (
+            <div
+              className={`absolute left-0 top-0 h-full rounded-full transition-all duration-300 ${
+                hasQuorum ? 'bg-green-500' : 'bg-blue-500'
+              }`}
+              style={{ width: `${(signedCount / validators.length) * 100}%` }}
+            />
+          )}
+        </div>
```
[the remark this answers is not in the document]
nambrot: Fixed in c2cc210 - added `validators.length > 0` guards around all division operations.

src/features/messages/cards/TimelineCard.tsx
```diff
@@ -1,25 +1,359 @@
...
+        <div
+          className="absolute left-0 top-0 h-full rounded-full bg-blue-500 transition-all duration-300"
+          style={{ width: `${(signedCount / validators.length) * 100}%` }}
+        />
+        <div
+          className="absolute top-0 h-full w-0.5 bg-gray-600"
+          style={{ left: `${(threshold / validators.length) * 100}%` }}
+          title={`Threshold: ${threshold}`}
+        />
+      </div>
```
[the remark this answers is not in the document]
nambrot: Fixed in c2cc210 - wrapped progress bar elements in `validators.length > 0` check.

src/features/messages/cards/TransactionCard.tsx
```diff
@@ -335,6 +347,56 @@ function DeliveryStatus({ children }: PropsWithChildren<unknown>) {
...
+        <p className="mt-2 text-center text-xs text-gray-500">Waiting for validators to sign...</p>
+      )}
+      {hasQuorum && (
+        <p className="mt-2 text-center text-xs text-green-600">
+          ✓ Quorum reached, waiting for relayer
+        </p>
+      )}
+    </div>
+  );
+}
```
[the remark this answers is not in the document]
nambrot: Fixed in c2cc210 - added `validators.length > 0` guard.

src/features/debugger/types.ts
```diff
@@ -9,17 +9,36 @@ export enum MessageDebugStatus {
   InvalidIsmDefinition = 'invalidIsmDefinition',
 }
 
+// Validator signature status for multisig ISMs
+export interface ValidatorStatus {
+  address: Address;
+  alias?: string;
+  status: 'signed' | 'pending' | 'error';
+  error?: string;
+}
```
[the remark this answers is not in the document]
nambrot: Fixed in c2cc210 - added `import type { Address, AddressTo, HexString } from '@hyperlane-xyz/utils';`

src/features/chains/useChainMetadata.ts
```diff
@@ -15,7 +10,9 @@ import { useQueryParam } from '../../utils/queryParams';
 
 const CHAIN_CONFIGS_KEY = 'chains';
 
-const ChainMetadataArraySchema = z.array(ChainMetadataSchema);
+// Use z.any() to avoid TypeScript infinite recursion with deep zod schemas
+// Runtime validation still happens via ChainMetadataSchema internally
+const ChainMetadataArraySchema = z.array(z.any());
```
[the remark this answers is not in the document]
nambrot: Fixed in c2cc210 - updated comment to accurately reflect that this skips validation of individual items.

.yarn/install-state.gz
paulbalaji, line 1: yarn?

paulbalaji: what happened to `git add -p` :P

src/pages/api/ism-details.ts
```diff
@@ -0,0 +1,305 @@
...
+import { config as appConfig } from '../../consts/config';
+import { logger } from '../../utils/logger';
+
+// Cache for registry, multiProvider and core
+let cachedRegistry: GithubRegistry | null = null;
+let cachedMultiProvider: MultiProvider | null = null;
+let cachedCore: HyperlaneCore | null = null;
+let cachedMetadataBuilder: BaseMetadataBuilder | null = null;
+let cacheTimestamp = 0;
+const CACHE_TTL = 60 * 1000; // 1 minute
```
Xaroz, line 30: iirc there is no guaranteed that in Vercel these global variables won't be re-instanced after each serverless/edge function requests

src/features/messages/cards/IsmDetailsCard.tsx
```diff
@@ -32,32 +40,306 @@ export function IsmDetailsCard({ ismDetails, blur }: Props) {
...
+  return (
+    <div className={`${depth > 0 ? 'ml-4 border-l-2 border-gray-200 pl-3' : ''}`}>
+      <div
+        className={`flex items-center space-x-2 py-1 ${hasChildren || isMultisig ? 'cursor-pointer' : ''}`}
+        onClick={() => (hasChildren || isMultisig) && setExpanded(!expanded)}
+      >
+        {/* Expand/collapse indicator */}
+        {(hasChildren || isMultisig) && (
+          <span className="w-3 text-xs text-gray-400">{expanded ? '▼' : '▶'}</span>
+        )}
```
Xaroz: `hasChildren` && isMultisig could be refactored in a single variable
`const isValid = hasChildren && isMultisig`

src/features/messages/cards/IsmDetailsCard.tsx
```diff
@@ -32,32 +40,306 @@ export function IsmDetailsCard({ ismDetails, blur }: Props) {
...
+          title={validator.error}
+        >
+          error
+        </span>
+      )}
+    </div>
+  );
+}
+
+// Helper functions
```
Xaroz: can we extract the helper functions below into its own file

src/features/messages/cards/IsmDetailsCard.tsx
```diff
@@ -32,32 +40,306 @@ export function IsmDetailsCard({ ismDetails, blur }: Props) {
...
+}
+
+function isAggregationResult(
+  result: MetadataBuildResult,
+): result is AggregationMetadataBuildResult {
+  return result.type === IsmType.AGGREGATION || result.type === IsmType.STORAGE_AGGREGATION;
+}
+
+function isRoutingResult(result: MetadataBuildResult): result is RoutingMetadataBuildResult {
+  return (
```
Xaroz: is the `result is ...BuildResult`, necessary here?

src/features/messages/cards/IsmDetailsCard.tsx
```diff
@@ -32,32 +40,306 @@ export function IsmDetailsCard({ ismDetails, blur }: Props) {
...
+          {/* Selected ISM for routing */}
+          {isRouting && (result as RoutingMetadataBuildResult).selectedIsm && (
+            <div className="space-y-1">
+              <span className="ml-6 text-xs text-gray-500">Routes to:</span>
+              <IsmTreeNode
+                result={(result as RoutingMetadataBuildResult).selectedIsm}
+                depth={depth + 1}
+              />
+            </div>
+          )}
```
Xaroz, line 140: I am not sure the casting is necessary here or in the other places, at least not repeated twice/multiple tines

src/features/messages/cards/TimelineCard.tsx
```diff
@@ -1,25 +1,345 @@
...
+  // 1. If status shows delivered, stage is Relayed
+  // 2. If quorum is reached from real-time validator data, stage is at least Validated
+  // 3. Otherwise use the stage from useMessageStage
+  let stage = _stage;
+  if (status === MessageStatus.Delivered) {
+    stage = MessageStage.Relayed;
+  } else if (hasQuorum && _stage < MessageStage.Validated) {
+    // Real-time validator data shows quorum reached, upgrade to Validated
+    stage = MessageStage.Validated;
+  }
```
Xaroz: Mutating variables with `let` and multiple `if` statements during render violates React's declarative philosophy—you're tracking state changes imperatively instead of deriving values directly, which makes the code harder to reason about, debug, and can cause issues if the logic ever needs to re-run. For simple cases use a ternary: `const stage = status === Delivered ? Relayed : hasQuorum ? Validated : _stage` .For complex logic like yours, use a utility function: extract to `deriveMessageStage(status, proposedStage, hasQuorum)` which returns the correct stage—this makes it testable, reusable, and keeps your render function purely declarative.

src/features/messages/cards/TimelineCard.tsx
```diff
@@ -1,25 +1,345 @@
...
+  );
+}
+
+function StageIcon({ stage }: { stage: MessageStage }) {
+  const iconMap: Partial<Record<MessageStage, string>> = {
+    [MessageStage.Sent]: '✈',
+    [MessageStage.Finalized]: '🔒',
+    [MessageStage.Validated]: '🛡',
+    [MessageStage.Relayed]: '✉',
+  };
```
Xaroz: can we use the actual icons that is in the current explorer instead of emojis

src/features/messages/cards/TimelineCard.tsx
```diff
@@ -1,25 +1,345 @@
...
+  );
+}
+
+function ChevronBlue() {
+  return (
+    <div className="absolute -right-3 top-0 h-6 w-3 overflow-hidden">
+      <div className="h-0 w-0 border-y-[12px] border-l-[12px] border-y-transparent border-l-blue-500" />
+    </div>
+  );
+}
```
Xaroz: there are chevrons components in the widget package

package.json
```diff
@@ -91,7 +91,8 @@
       "cipher-base": "1.0.5",
       "elliptic": "6.6.1",
       "pbkdf2": "3.1.3",
-      "form-data": "4.0.4"
+      "form-data": "4.0.4",
+      "@hyperlane-xyz/sdk": "21.1.0-preview.9c72f6fb2deb13238aa68a4a83bff7a7156047e6"
```
Xaroz: remove this

src/features/debugger/types.ts
```diff
@@ -1,3 +1,5 @@
+import type { Address, AddressTo, HexString } from '@hyperlane-xyz/utils';
```
Xaroz: type error here it seems

.yarn/install-state.gz
paulbalaji: Removed - this was accidentally committed. Rebuilt the branch cleanly from main in f3a9d01.

src/pages/api/ism-details.ts
```diff
@@ -0,0 +1,305 @@
...
+        ismConfig: ismFromCache,
+        hookConfig: hookFromCache,
+      };
+    }
+
+    return res.status(200).json(response);
+  } catch (error: any) {
+    logger.error('Error fetching ISM details:', error);
+    logger.error('[TIMING] Error occurred after:', timer.getTimings());
+    return res.status(500).json({ error: error.message || 'Internal server error' });
```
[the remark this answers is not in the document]
paulbalaji: Fixed in f3a9d01 - now returns generic `'Internal server error'` instead of raw `error.message`. Full error is still logged server-side.

src/pages/api/ism-details.ts
paulbalaji: Good call - added a comment acknowledging this is best-effort caching that only helps within a single warm instance. The caches still provide value when the same instance handles multiple requests, but we're not relying on it for correctness.

src/features/messages/cards/IsmDetailsCard.tsx
paulbalaji: Done - extracted into `const isExpandable = isAggregation || isRouting || isMultisig` and use it for the cursor, click handler, and expand indicator.

src/features/messages/cards/IsmDetailsCard.tsx
paulbalaji: Done - extracted all helpers into `src/features/messages/cards/ismHelpers.ts` (type guards, display helpers, and `extractValidatorInfo`).

src/features/messages/cards/IsmDetailsCard.tsx
paulbalaji: Yes - the `result is ...BuildResult` type predicate is needed for TypeScript type narrowing. Without it, TS won't narrow the type after the guard check, so you'd need `as` casts everywhere. With the type guard, code after `if (isMultisigResult(result))` knows `result` is a `MultisigMetadataBuildResult`.

src/features/messages/cards/IsmDetailsCard.tsx
paulbalaji: Agreed the repeated casts are verbose. The type guards narrow within `if` blocks, but JSX expressions need explicit casts. Kept as-is for now since it's consistent - could revisit with discriminated unions.

src/features/messages/cards/TimelineCard.tsx
paulbalaji: Fixed - extracted to `deriveMessageStage(status, proposedStage, hasQuorum)` utility function. Render is now purely declarative with `const stage = deriveMessageStage(...)`.

src/features/messages/cards/TimelineCard.tsx
paulbalaji: Fixed - now using `AirplaneIcon`, `LockIcon`, `ShieldIcon`, `EnvelopeIcon` from `@hyperlane-xyz/widgets` instead of emojis.

src/features/messages/cards/TimelineCard.tsx
paulbalaji: Fixed - replaced custom CSS chevrons with `WideChevronIcon` from `@hyperlane-xyz/widgets`.

package.json
paulbalaji: Removed - rebuilt the branch from scratch on latest main. No more preview SDK resolutions in package.json.

src/features/debugger/types.ts
paulbalaji: Fixed - the `AddressTo` import doesn't exist in the current SDK version. Rebuilt from main's types.ts which uses global `Address`/`HexString`/`AddressTo` types without explicit import.

paulbalaji:
## Review Summary

Consolidated review from automated tools and manual inspection. Most prior CodeRabbit/Xaroz feedback has been addressed — this covers remaining issues.

### High

| # | Issue | File |
|---|-------|------|
| 1 | **Hook cache key missing destination** — routing/fallback hooks resolve differently per destination, so `${originChain}:${address}` can serve a poisoned cache entry for a different destination | `ism-details.ts:48` |
| 2 | **`extractValidatorInfo` returns only first multisig branch** — for aggregation/routing ISMs with multiple multisig children, one branch reaching quorum ≠ full ISM buildable. UI can show "Validated" / "quorum reached" too early | `ismHelpers.ts:93` |

### Medium

| # | Issue | File |
|---|-------|------|
| 3 | **`z.any()` removes runtime validation** of user-controlled `chains` query param. Malformed links (`[null]`, `["x"]`) now pass `safeParse` and throw at `c.name` dereference | `useChainMetadata.ts:14` |
| 4 | **No EVM gate on `useIsmDetails`** — called for any found message including PI/non-EVM. The API assumes EVM receipt shape (`getTransactionReceipt`, `getDispatchedMessages`), so non-EVM detail pages issue a failing API request + server error on every load | `MessageDetailsInner.tsx:105` |
| 5 | **Missing SDK exports** — `MetadataBuildResult`, `BaseMetadataBuilder`, `ValidatorInfo`, etc. don't exist in SDK 31.1.0. CI typecheck/build fails | multiple files |

### Low

| # | Issue | File |
|---|-------|------|
| 6 | **`hasQuorum` inconsistency** — `ValidatorList` uses `signedCount >= threshold` without `threshold > 0` guard (other components have it) | `IsmDetailsCard.tsx:155` |
| 7 | **Sequential `getChainAddresses`** — awaits each chain serially in a loop, impacts cold-start time | `ism-details.ts:116-125` |
| 8 | **`originDomain` validated late** — after `originTxHash`/`messageId` already processed | `ism-details.ts:167-179` |
| 9 | **Format failures** on `TimelineCard.tsx` and `ism-details.ts` | CI |

Working on fixes now.

PR #294, opened 2026-03-16
perf: improve page load and rendering performance

Xaroz:
### Rendering
- **GPU-accelerate background overlay** — `will-change: transform` + `translateZ(0)` on the grid overlay to eliminate scroll repaints from mask compositing
- **Memoize hot components** — `KeyValueRow` (100+ per detail page), `MessageSummaryRow` (40-100 per table)
- **Replace slow wildcard CSS selectors** — `[class*='htw-']` attribute selectors replaced with specific class selectors for widget overrides
- **Reduce accent-glow blur** from 20px to 10px

### Loading
- **Preload custom fonts** via `<link rel="preload">` to reduce FOUT
- **Parallelize block explorer URL lookups** in ContentDetailsCard (was sequential awaits)
- **Enable Link prefetch** on message table rows

## Test plan

- [ ] Verify background grid overlay scrolls smoothly
- [ ] Verify widget styling overrides still apply (chain picker, modals, tooltips)
- [ ] Verify fonts load without visible flash
- [ ] Visual check that accent-glow shadow still looks good at 10px
- [ ] OG card previews still work (no SSR changes in this PR)

### DIFF
--- a/src/pages/message/[messageId].tsx
+++ b/src/pages/message/[messageId].tsx
@@ -100,7 +100,10 @@ export const getServerSideProps: GetServerSideProps<PageProps> = async (ctx) => 
 
   // Skip OG data fetching for non-bot user agents (speeds up navigation for real users)
   const userAgent = ctx.req?.headers['user-agent'] || '';
-  const isBot = /bot|crawl|spider|slurp|facebookexternalhit|Twitterbot|LinkedInBot|Discordbot|WhatsApp/i.test(userAgent);
+  const isBot =
+    /bot|crawl|spider|slurp|facebookexternalhit|Twitterbot|LinkedInBot|Discordbot|WhatsApp/i.test(
+      userAgent,
+    );
 
   if (isBot && messageId && typeof messageId === 'string') {
     try {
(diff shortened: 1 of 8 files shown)


src/pages/message/[messageId].tsx
```diff
@@ -125,6 +132,11 @@ export const getServerSideProps: GetServerSideProps<PageProps> = async (ctx) =>
     }
   }
 
+  // Cache the response for real users (instant navigation) while bots get fresh data
+  if (!isBot) {
+    ctx.res.setHeader('Cache-Control', 'public, s-maxage=60, stale-while-revalidate=300');
```
Comment (paulbalaji) [inline src/pages/message/[messageId].tsx:137]: `public, s-maxage=60` on the non-bot HTML is cache-unsafe here. Because the response varies by user agent but doesn’t have a safe cache key split, a CDN can cache the `ogData: null` human response and serve it to bots, breaking OG tags in the common user-visit-then-share flow. In `#295` this is fixed by avoiding shared HTML caching on this route rather than relying on bot/human cache variation.

PR #301, opened 2026-03-25
perf: keep explorer metadata-first and evm-only at runtime

paulbalaji:
## Summary

- rebased on latest `main`
- updated to released packages:
  - `@hyperlane-xyz/core@11.3.1`
  - `@hyperlane-xyz/registry@24.3.0`
  - `@hyperlane-xyz/sdk@33.0.2`
  - `@hyperlane-xyz/utils@33.0.2`
  - `@hyperlane-xyz/widgets@33.0.2`
- keeps Explorer metadata-first and runtime/provider work narrow at runtime
- now uses the released narrow EVM-like runtime surface from `hyperlane-xyz/hyperlane-monorepo#8641`, preserving Tron parity for the fee path without widening back to the all-VM provider-builder surface
- cuts back to the shared widgets `ChainSearchMenu` surface instead of the temporary local picker implementation
- includes the merged `ChainMetadataResolver` duplicate-`chainId` fix, so Explorer no longer crashes on the current registry snapshot

## What changed

1. Explorer runtime/provider setup stays metadata-first and narrow at runtime.
2. metadata/read helpers come from monorepo narrow surfaces instead of local copies.
3. fee/runtime provider wiring now uses the released `runtime/evmLike` surface, preserving the existing Ethereum + Tron fee behavior without pulling every VM builder back into the graph.
4. provider lifecycle/readiness is explicit so runtime-dependent queries wait for a ready provider and invalidate correctly when it changes.
5. chain search is back on the shared widgets `ChainSearchMenu` path rather than a temporary local reinvention.

## Canonical local benchmark

Measured serially, one build at a time, from fresh temp worktrees with:
- `pnpm install`
- `pnpm clean`
- `/usr/bin/time -p pnpm build`

| State | Build time | Compile time | `.next` total | `.next/static/chunks` | static chunk JS bytes | `.next/server` |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| current `main` (`b79ab39`) | `39.63s` | `33.0s` | `2,285,172 KB` | `44,960 KB` | `45,937,245` | `44,748 KB` |
| this branch (`840e1e5`) | `18.59s` | `15.7s` | `788,756 KB` | `56,392 KB` | `57,462,087` | `174,920 KB` |

Interpretation:
- still clearly faster than latest `main` on clean local build time
- overall emitted `.next` output is much smaller than latest `main`
- some individual emitted buckets are larger than `main`, but the branch still lands materially smaller overall

## Validation

- [x] `pnpm typecheck`
- [x] `pnpm lint`
- [x] `pnpm build`
- [x] `pnpm test src/features/messages/warpFees/fetchWarpFees.test.ts --runInBand`
- [x] `pnpm test src/utils/warpRouteAmounts.test.ts`
- [x] resolver construction no longer throws on duplicate `chainId`

## Notes

- invalid non-null warp-route `scale` values now warn before falling back to `1:1`
- provider readiness stays stable while rebuilding from metadata when a last-known-good provider already exists

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes provider construction/readiness and several runtime-dependent queries (PI search, delivery status, balances/fees) now depend on a new async EVM-like runtime adapter and cache invalidation behavior.
> 
> **Overview**
> Switches explorer runtime/provider wiring from `MultiProtocolProvider` to a new async `ExplorerMultiProvider` built via `runtime/evmLike` (`createRuntimeMultiProvider`), and updates PI search and API helpers to construct providers through this narrow runtime path.
> 
> Introduces explicit provider readiness and a `multiProviderVersion` in `store.ts`, and updates React Query keys/enabled flags across delivery status, PI queries, warp fees/balances, collateral, and rebalances so runtime-backed reads wait for a ready provider and invalidate when it rebuilds.
> 
> Removes the local `metadataManager` implementation/tests in favor of SDK `ChainMetadataResolver` and moves many local warp-route/token types to `@hyperlane-xyz/sdk/warp/read`; collateral/balance fetching is tightened to EVM-only via `createEvmHypAdapter`. Also updates `@hyperlane-xyz/core` to `11.3.1`, adjusts Next route typings import, and adds a small CSS `z-index` override (`.htw-z-30`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 396ddff2204b663842e8c733d40cd848ce88af0e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/features/messages/collateral/useCollateralStatus.ts
```diff
@@ -90,15 +105,15 @@ async function fetchCollateralBalance(
...
 
     logger.debug('Fetched collateral balance', {
       chain: destinationToken.chainName,
       token: destinationToken.symbol,
-      balance: tokenAmount.amount.toString(),
+      balance: tokenAmount.toString(),
     });
 
-    return tokenAmount.amount;
+    return tokenAmount;
```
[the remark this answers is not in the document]
paulbalaji: Fixed on the current branch.

`useCollateralStatus` and `useActiveRebalances` now both gate on `useReadyMultiProvider()`, include `multiProviderVersion` in the query key, and only execute once the runtime provider is ready, so empty-provider results are no longer cached for the stale window.

src/features/messages/warpVisualization/useWarpRouteBalances.ts
```diff
@@ -1,21 +1,15 @@
...
-import { ProtocolType } from '@hyperlane-xyz/utils';
+} from '@hyperlane-xyz/sdk/token/adapters/EvmTokenAdapter';
+import { TokenStandard } from '@hyperlane-xyz/sdk/token/TokenStandard';
 import { useQuery } from '@tanstack/react-query';
 import { useMemo } from 'react';
 
-import { useMultiProvider } from '../../../store';
+import { useMultiProviderVersion, useReadyMultiProvider } from '../../../store';
 import { logger } from '../../../utils/logger';
+import type { ExplorerMultiProvider as MultiProtocolProvider } from '../../hyperlane/sdkRuntime';
```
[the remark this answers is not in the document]
paulbalaji: Checked this against the current SDK package exports before changing anything.

These adapter imports are using public exported SDK subpaths (`@hyperlane-xyz/sdk/package.json` exports `./token/*`), not private file-path imports. Keeping the narrow subpath imports here is intentional so this page does not widen back to the SDK root barrel on the hot path. I left `TokenStandard` on the narrow token subpath for the same reason.

src/features/messages/collateral/useCollateralStatus.ts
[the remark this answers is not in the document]
paulbalaji: Fixed in `5c2fb6b`.

On rebuild failure we now preserve `isMultiProviderReady` when there is already a last-known-good provider in the store, instead of leaving the readiness flag stuck false. That keeps RPC-dependent features enabled on the previous provider while still allowing future resync attempts.

src/features/messages/warpVisualization/useWarpRouteBalances.ts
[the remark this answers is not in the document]
paulbalaji: No code change here.

`loadWarpRouteData(...)` was only used inside `metadataStore.ts`, so I dropped the unnecessary export in `5c2fb6b` to keep the module surface local.

Xaroz:
@claude review no cap

src/features/chains/ChainSearchModal.tsx
```diff
@@ -1,9 +1,36 @@
...
+    family: etherscan
+    url: https://sepolia.etherscan.io
+    apiUrl: https://api-sepolia.etherscan.io/api
+    apiKey: '12345'
+blocks:
+  confirmations: 1
+  estimateBlockTime: 13
+`;
+
 export function ChainSearchModal({
```
Xaroz, line 7: I think there is room to increase the height and width of the modal, as you can see it also does horizontal scrolling which we wanna prevent

Image

src/features/chains/ChainSearchModal.tsx
```diff
@@ -15,25 +42,291 @@ export function ChainSearchModal({
...
+        chainName !== metadata.name && getEffectiveDomainId(chain) === effectiveDomainId,
+    )
+  ) {
+    return { success: false, error: 'domainId is already in use by another chain' };
+  }
+
+  return { success: true, data: metadata };
+}
+
+function areChainIdsEqual(
```
Xaroz: can these be utils function be moved out the component instead?

src/features/chains/ChainSearchModal.tsx
Xaroz: We are also missing filtering, sorting and editing a chain with this new modal

src/features/messages/collateral/useCollateralStatus.ts
```diff
@@ -52,19 +57,29 @@ async function fetchCollateralBalance(
       return undefined;
     }
 
-    // Create Token instance from the destination token config
```
Xaroz, line 55: not sure if theres an early check before fetchCollateralBalance but maybe we can do an early return if protocol !== ethereum?

src/features/chains/ChainSearchModal.tsx
paulbalaji: Width/overflow is fixed on this branch in `9c38056`. I’m not going to add filter/sort/edit to the temporary local modal though, because that just duplicates the widget surface we want to get back to. I opened `hyperlane-xyz/hyperlane-monorepo#8637` to make `ChainSearchMenu` metadata-first friendly instead, and `#301` is now explicitly blocked on that follow-up so this local modal can be dropped again after release.

src/features/chains/ChainSearchModal.tsx
paulbalaji: These helpers are already module-scope, not nested inside the component. I’m leaving them local to this temporary file for now because the longer-term direction is to drop this file once `hyperlane-xyz/hyperlane-monorepo#8637` lands and Explorer cuts back to the shared widgets chain menu path.

src/features/messages/collateral/useCollateralStatus.ts
paulbalaji: Fixed in `5e36845`. `fetchCollateralBalance(...)` now exits early once chain metadata is available and the destination protocol is not EVM, so we skip the rest of the adapter/provider path for non-EVM chains.

src/utils/warpRouteAmounts.test.ts
```diff
@@ -44,20 +47,44 @@ describe('getWarpRouteAmountParts', () => {
...
-  it('handles negative scale (assertValidScale throws)', () => {
+  it('handles invalid rational scales as invalid (falls back to 1)', () => {
     const messageAmount = 1_000_000n;
-    expect(() => getWarpRouteAmountParts(messageAmount, { decimals: 6, scale: -10 })).toThrow();
+    const result = getWarpRouteAmountParts(messageAmount, {
+      decimals: 6,
+      scale: { numerator: 1.5, denominator: 1 },
+    });
+    expect(result).toEqual({ amount: 1_000_000n, decimals: 6 });
   });
```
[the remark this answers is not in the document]
paulbalaji: Added the missing object-scale edge coverage in `5e36845`: `NaN`, `Infinity`, zero denominator, and negative denominator all now have explicit tests around the fallback path.

src/features/hyperlane/sdkRuntime.ts
```diff
@@ -0,0 +1,21 @@
+import type { ChainMetadata } from '@hyperlane-xyz/sdk/metadata/chainMetadataTypes';
+import { MultiProviderAdapter } from '@hyperlane-xyz/sdk/providers/MultiProviderAdapter';
+import type { ChainMap } from '@hyperlane-xyz/sdk/types';
+
```
paulbalaji, line 4: **Follow-up with #290:** once this lands, `src/features/messages/warpFees/fetchWarpFees.ts` should be re-typed against `ExplorerMultiProvider` instead of importing `MultiProtocolProvider` from `@hyperlane-xyz/sdk`, so the adapter is the single provider type used across the explorer.

src/features/hyperlane/sdkRuntime.ts
```diff
@@ -0,0 +1,21 @@
...
+// Explorer's provider-backed runtime paths are EVM-only today:
+// PI search, delivery status, debugger, ICA, rebalances, and warp balance reads.
+// Keeping this helper Ethereum-only avoids pulling every VM builder into the build graph.
+export async function createRuntimeMultiProvider(
+  chainMetadata: ChainMap<ChainMetadata<{ mailbox?: string }>>,
+): Promise<ExplorerMultiProvider> {
+  const { evmRuntimeProviderBuilders } = await import('@hyperlane-xyz/sdk/providers/runtime/evm');
+  return new MultiProviderAdapter(chainMetadata, {
+    providerBuilders: evmRuntimeProviderBuilders,
+  });
```
paulbalaji, line 22: **Tron compatibility question for #290:** #290's `fetchWarpFees` calls `multiProvider.getTronProvider()` as an `isEVMLike` path. Registering only `evmRuntimeProviderBuilders` here means the Tron branch will silently return nothing under `MultiProviderAdapter` once #290 starts routing through this. Either add a Tron builder here, or tighten #290's `isEVMLike` gate to Ethereum-only until a Tron builder exists — worth deciding before merge so Tron fee-display doesn't regress.

paulbalaji (commented):
## Follow-ups to reconcile with #290 (warp-fee parsing)

Drive-by observations, not a blocking review. #290 landed new runtime-reading code (`useWarpFees` / `fetchWarpFees`) that predates the patterns this PR establishes. When this PR merges (or in a fast follow-up), a few small sweeps would be worth doing to keep 290's code aligned with the new conventions:

**1. Migrate `useWarpFees` to the readiness-gated hooks.** `src/features/messages/warpFees/useWarpFees.ts:11` currently calls `useMultiProvider()` — the pre-301 pattern. Under this PR it should switch to `useReadyMultiProvider()` + include `useMultiProviderVersion()` in the `queryKey`, matching what's being done here for `useMessageDeliveryStatus`, `useActiveRebalances`, `useCollateralStatus`, `useWarpRouteBalances`, and `usePiChainMessageSearchQuery`. Otherwise `fetchWarpFees` can fire against the empty/old `MultiProviderAdapter` and cache the stale result.

**2. Re-type `fetchWarpFees` against `ExplorerMultiProvider`.** `src/features/messages/warpFees/fetchWarpFees.ts` imports `MultiProtocolProvider` directly from `@hyperlane-xyz/sdk`. After this lands, that should become `ExplorerMultiProvider` from `../hyperlane/sdkRuntime` so the adapter surface is the one type used throughout the explorer.

**3. Tron path under the new adapter.** `fetchWarpFees` calls `multiProvider.getEthersV5Provider()` and `multiProvider.getTronProvider()`. This PR only registers `evmRuntimeProviderBuilders` in `createRuntimeMultiProvider`, so any Tron warp transfer will fall off the `MultiProviderAdapter` once this lands. Either register a Tron builder in `sdkRuntime.ts`, or tighten #290's `isEVMLike` gate to Ethereum-only until Tron builders land. Worth a conscious decision either way before this merges so Tron fee-display doesn't silently regress.

**4. Root-import sweep.** Per the codex follow-up notes, a few root imports still exist where narrow subpaths are available — `TokenArgs` in `WarpTransferDetailsCard.tsx`, `ChainMetadata` / `ChainMap`, `TokenStandard`, `WarpCoreConfig`. Mostly type-only, so low leverage, but a natural extension of the narrow-surface work in this PR. (Skip `scalesEqual`, `ScaleInput`, `isProxy`, `proxyImplementation` — hyperlane-monorepo#8433 did not expose lighter public surfaces for those.)

**5. `PROTOCOL_TO_HYP_NATIVE_STANDARD`.** `fetchWarpFees` imports this from `@hyperlane-xyz/sdk` root. If a narrow subpath exists (e.g. `./token/...`) it should move there for consistency with the rest of the adapter imports in this PR.

All five are post-merge tidy-ups except #3, which is worth confirming before merge since it's behavioural.

src/store.ts
```diff
@@ -88,7 +116,14 @@ export function useMultiProvider() {
...
-  const multiProvider = useMultiProvider();
-  if (!multiProvider.getKnownChainNames().length) return undefined;
+  const multiProvider = useProviderStore((s) => s.multiProvider);
+  const isMultiProviderReady = useProviderStore((s) => s.isMultiProviderReady);
+  ensureProviderStoreSubscription();
+  if (!isMultiProviderReady || !multiProvider.getKnownChainNames().length) return undefined;
   return multiProvider;
 }
+
+export function useMultiProviderVersion() {
```
paulbalaji, line 126: **Follow-up with #290:** `useWarpFees` in #290 still uses `useMultiProvider()`. After this merges it should migrate to `useReadyMultiProvider()` + include `useMultiProviderVersion()` in its query key, same pattern as the other runtime hooks in this PR. Otherwise the fee query can fire against the empty adapter pre-ready and cache a stale / null result.

src/features/chains/ChainSearchModal.tsx
```diff
@@ -15,22 +15,24 @@ export function ChainSearchModal({
...
+  const setChainMetadataOverrides = useStore((s) => s.setChainMetadataOverrides);
 
-  const onClick = onClickChain || (() => {});
+  const handleClickChain = (metadata: ChainMetadata) => {
+    if (!onClickChain) return;
+    onClickChain(metadata);
+    close();
+  };
 
   return (
```
Xaroz, line 29: this is happening live too but I think we need to fix the z-index for the filter 

Image

src/utils/warpRouteAmounts.ts
```diff
@@ -114,8 +112,14 @@ export function getWarpRouteAmountParts(
...
   const tokenDecimals = decimals ?? DEFAULT_TOKEN_DECIMALS;
+  const parsedScale = parseScale(scale);
+  if (scale != null && !parsedScale) {
+    logger.warn('Invalid warp route scale; falling back to 1:1', { scale });
+  }
+  const scaleValue = parsedScale ?? { numerator: 1n, denominator: 1n };
+  // bigint division truncates toward zero (floor for positive token amounts)
   return {
-    amount: localAmountFromScale(messageAmount, scale),
+    amount: (messageAmount * scaleValue.denominator) / scaleValue.numerator,
```
Xaroz: why did we remove the utils fn?

src/utils/warpRouteAmounts.ts
```diff
@@ -25,37 +25,35 @@ export interface EffectiveDecimalsToken {
...
+  if (!Number.isFinite(value) || !Number.isInteger(value)) return null;
+  if (!Number.isSafeInteger(value)) return null;
+  return BigInt(value);
 }
 
-function assertValidScale(scale: NormalizedScale): void {
-  if (scale.numerator <= 0n || scale.denominator <= 0n) {
-    throw new Error(`Scale must be positive, got ${scale.numerator}/${scale.denominator}`);
+function parseScale(scale: WarpRouteAmountConfig['scale']): ParsedScale | null {
+  if (scale === undefined || scale === null) return null;
```
Xaroz, line 38: still can't use utils here right? (isNullish)

Xaroz (approved):

Xaroz:
@claude wanna review again buddy with the new changes

src/features/hyperlane/sdkRuntime.ts
paulbalaji: Handled on the follow-up path now. `fetchWarpFees.ts` is already typed against `ExplorerMultiProvider`, so the explorer stays on the adapter type instead of reaching back to the SDK `MultiProtocolProvider`.

src/features/hyperlane/sdkRuntime.ts
paulbalaji: Handled now too. Explorer is on the released narrow evm-like runtime builder path, so the Tron fee branch is preserved without widening back to the broad provider-builder surface.

src/store.ts
paulbalaji: Handled in the current branch. `useWarpFees` now uses `useReadyMultiProvider()` and `useMultiProviderVersion()` in the query key, same readiness/versioning pattern as the other provider-backed hooks.

src/features/chains/ChainSearchModal.tsx
paulbalaji: Fixed locally here too. `ChainSearchModal` now applies an Explorer-side z-index override on the shared widgets filter/dropdown layers inside the modal shell, so the filter panel stacks above the results list instead of getting buried under it.

src/utils/warpRouteAmounts.ts
paulbalaji: Put the tiny helper back, but kept the newer structure. `localAmountFromScale(...)` is restored so the amount conversion has a named unit again, while `parseScale(...)` + `toIntegerBigInt(...)` still handle the validation/normalization path.

src/utils/warpRouteAmounts.ts
paulbalaji: Still avoiding `@hyperlane-xyz/utils` here. `warpRouteAmounts.ts` is imported by the OG edge route, so I kept the explicit null checks instead of widening that path just for `isNullish(...)`.

PR #321, opened 2026-04-27
fix(tron): match warp-route entries via canonical bytes32

paulbalaji:
## Summary

Tron-destination rows on the explorer message list (e.g. `?destination=tron`) silently dropped the warp transfer label (USDT amount + symbol). Tron messages still appeared, but without the token icon/amount cell.

### Root cause

`getTokenFromWarpRouteChainAddressMap` (`src/utils/token.ts`) compared the resolved recipient against registry keys via `tokenAddress.endsWith(address)`. The two sides are in different encodings for Tron:

- **Registry keys**: hyperlane-registry stores Tron warp-route entries as EVM-shaped hex (e.g. `0xbf8078818627110fD05827Ca0aa9E4518d3421ec` — verified in `deployments/warp_routes/USDT/eclipsemainnet-config.yaml` under `chainName: tron`). The SDK's `buildWarpRouteMaps` keys the per-chain map by `addressOrDenom` literally.
- **Resolved recipient**: by the time `parseWarpRouteMessageDetails` runs, `postgresByteaToAddress` (`src/features/messages/queries/encoding.ts`) has already decoded the wire bytes to **Tron base58** using the destination chain's metadata.

`endsWith` of a hex registry key against a base58 recipient never matches → `parseWarpRouteMessageDetails` returns `undefined` → `MessageTable` skips the warp cell.

### Note on the previous commit

The previous attempt (bcc534d) added a Tron branch to `formatAddress` that converted `bytes32 → base58`. In production this branch was a no-op: recipients reaching `formatAddress` had already been decoded to base58, so the new branch hit `isAddressTron(address) → true` and short-circuited. The registry-vs-recipient mismatch was unchanged.

That commit is reverted in this PR; the proper fix lives in the comparison logic itself.

## Fix

Replace the `endsWith` match in `getTokenFromWarpRouteChainAddressMap` with a canonical-form comparison:

- Normalize each side to lowercase bytes32 hex via `addressToBytes32`.
- Route EVM-shaped hex through the EVM decoder (handles both genuine EVM addresses and Tron entries the registry stores in hex form).
- Route base58 / bech32 / etc. through the chain's protocol decoder.
- Fall back to legacy `endsWith` only when a registry key can't be canonicalized (e.g. IBC denoms).

Both forms of the same Tron address resolve to the same 20-byte payload → match.

Example transfer that should now display USDT in the list: https://explorer.hyperlane.xyz/message/0xd29637bc06874c59c6af1a786dbca307d38d43475f482d312a5bb6c38951c05c

### Out of scope (separate Tron gaps, not in this PR)

- `useCollateralStatus.ts` and `useWarpRouteBalances.ts` are still gated on `ProtocolType.Ethereum` only (no balance/collateral panels on Tron destination message detail pages). Worth a follow-up.

## Test plan

- [x] `pnpm run test -- src/features/messages/utils.test.ts` — 6/6 pass. The Tron regression test now mirrors production shape: hex registry key (`0xbf80…3421ec`) + base58 recipient.
- [x] `pnpm run typecheck` — clean.
- [x] `pnpm run lint` — clean.
- [ ] Manual: visit `/?destination=tron` against the live Hasura endpoint and confirm the USDT amount cell renders for known Tron transfers.
- [ ] Manual: open a Tron-destination message detail page and confirm the recipient field shows the `T…` base58 form.

> Note: 8 pre-existing PI-chain network test failures are unrelated to this change.

---

> [!NOTE]
> **Medium Risk**
> Changes the core token lookup logic used to label warp transfers, so regressions could cause missing/incorrect token matches across chains or denom formats despite a conservative fallback.
> 
> **Overview**
> Fixes warp-route token resolution by replacing the previous `endsWith`-based match with a canonical comparison that normalizes both registry keys and message addresses to lowercase `bytes32` (treating EVM-shaped hex as Ethereum even on non-EVM chains, e.g. Tron).
> 
> Adds targeted tests covering Tron hex-registry-key vs base58-recipient matching (origin and destination) and ensures behavior remains correct for checksum-cased EVM addresses and non-address denom keys via a suffix fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 90da4716e8bc75b39b9cc02d15da094256cb1654. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/features/messages/utils.test.ts
+++ b/src/features/messages/utils.test.ts
@@ -1,7 +1,12 @@
 import { type ChainMetadata, TokenStandard } from '@hyperlane-xyz/sdk';
 import type { ChainMetadataResolver } from '@hyperlane-xyz/sdk/metadata/ChainMetadataResolver';
 import type { WarpRouteChainAddressMap } from '@hyperlane-xyz/sdk/warp/read';
-import { addressToBytes32, ProtocolType } from '@hyperlane-xyz/utils';
+import {
+  addressToBytes32,
+  bytesToProtocolAddress,
+  fromHexString,
+  ProtocolType,
+} from '@hyperlane-xyz/utils';
 
 import { MessageStatus, type MessageStub } from '../../types';
 import { parseWarpRouteMessageDetails } from './utils';
@@ -183,16 +188,20 @@ describe('parseWarpRouteMessageDetails', () => {
       expect(result!.destAmount).toBe('1');
     });
 
-    it('matches a Tron destination token whose registry address is base58 (T...)', () => {
-      // USDT TRC20 on Tron — registry stores the base58 form, but the wire
-      // recipient is bytes32 hex. The fix in formatAddress converts the wire
-      // recipient back to base58 so the warp route lookup can match.
-      const TRON_USDT_BASE58 = 'TR7NHqjeKQxGTCi8q8ZY4pL8otSzgjLj6t';
-      const TRON_USDT_BYTES32 = addressToBytes32(TRON_USDT_BASE58, ProtocolType.Tron);
+    it('matches a Tron destination token whose registry key is EVM-shaped hex', () => {
+      // Production shape: hyperlane-registry stores Tron warp-route addresses as
+      // EVM-style hex (e.g. 0xbf80...3421ec), while postgresByteaToAddress decodes
+      // the wire recipient to Tron base58 before parseWarpRouteMessageDetails sees it.
+      const TRON_ROUTER_HEX = '0xbf8078818627110fD05827Ca0aa9E4518d3421ec';
+      const TRON_ROUTER_BYTES32 = addressToBytes32(TRON_ROUTER_HEX, ProtocolType.Ethereum);
+      const TRON_ROUTER_BASE58 = bytesToProtocolAddress(
+        fromHexString(TRON_ROUTER_BYTES32),
+        ProtocolType.Tron,
+      );
 
       const messageAmount = 10n ** 6n;
       const amountHex = messageAmount.toString(16).padStart(64, '0');
-      const recipientHex = TRON_USDT_BYTES32.slice(2);
+      const recipientHex = TRON_ROUTER_BYTES32.slice(2);
       const body = '0x' + recipientHex + amountHex;
 
       const message: MessageStub = {
@@ -201,12 +210,12 @@ describe('parseWarpRouteMessageDetails', () => {
         msgId: '0xabc',
         nonce: 1,
         sender: SENDER_BYTES32,
-        recipient: TRON_USDT_BYTES32,
+        recipient: TRON_ROUTER_BASE58,
         originChainId: 1,
         originDomainId: ORIGIN_DOMAIN,
         destinationChainId: 728126428,
         destinationDomainId: DEST_DOMAIN,
-        origin: { timestamp: 0, hash: '0x0', from: SENDER, to: TRON_USDT_BASE58 },
+        origin: { timestamp: 0, hash: '0x0', from: SENDER, to: TRON_ROUTER_BASE58 },
         body,
       };
 
@@ -223,10 +232,10 @@ describe('parseWarpRouteMessageDetails', () => {
           },
         },
         tron: {
-          [TRON_USDT_BASE58]: {
+          [TRON_ROUTER_HEX]: {
             chainName: 'tron',
             standard: TokenStandard.EvmHypSynthetic,
-            addressOrDenom: TRON_USDT_BASE58,
+            addressOrDenom: TRON_ROUTER_HEX,
             decimals: 6,
             symbol: 'USDT',
             name: 'Tether USD',
@@ -262,8 +271,9 @@ describe('parseWarpRouteMessageDetails', () => {
       );
 
       expect(result).toBeDefined();
+      expect(result!.originToken.symbol).toBe('USDT');
       expect(result!.destinationToken.symbol).toBe('USDT');
-      expect(result!.transferRecipient).toBe(TRON_USDT_BASE58);
+      expect(result!.transferRecipient).toBe(TRON_ROUTER_BASE58);
     });
 
     it('computes destAmount when only dest has scale (same decimals)', () => {
--- a/src/features/messages/utils.test.ts
+++ b/src/features/messages/utils.test.ts
@@ -1,7 +1,7 @@
 import { type ChainMetadata, TokenStandard } from '@hyperlane-xyz/sdk';
 import type { ChainMetadataResolver } from '@hyperlane-xyz/sdk/metadata/ChainMetadataResolver';
 import type { WarpRouteChainAddressMap } from '@hyperlane-xyz/sdk/warp/read';
-import { ProtocolType } from '@hyperlane-xyz/utils';
+import { addressToBytes32, ProtocolType } from '@hyperlane-xyz/utils';
 
 import { MessageStatus, type MessageStub } from '../../types';
 import { parseWarpRouteMessageDetails } from './utils';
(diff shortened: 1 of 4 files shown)


src/utils/addresses.ts
```diff
@@ -19,6 +22,11 @@ export function formatAddress(
   switch (metadata?.protocol) {
     case ProtocolType.Radix:
       return hexToRadixCustomPrefix(address, 'component', metadata?.bech32Prefix, 30);
+    case ProtocolType.Tron:
+      // Wire format is bytes32 hex; registry stores Tron token addresses as base58.
+      // Convert so warp route lookups (endsWith match) and UI both use base58.
+      if (isAddressTron(address)) return address;
+      return bytesToProtocolAddress(fromHexString(address), ProtocolType.Tron);
```
paulbalaji: This normalizes Tron recipients to base58, but the live `warpRouteChainAddressMap.tron` from `hyperlane-registry@main` is keyed by hex router addresses (`0xbf807881...`, `0x43Bb7C...`), not `T...` base58. Production query parsing already converts the DB `recipient` to base58 (`TTRmy...` for `0xbf807...`), so `getTokenFromWarpRouteChainAddressMap()` still compares hex map keys against a base58 lookup value and `endsWith()` never matches. This matches the preview behavior: Tron rows still have an empty Warped Token cell. The fix needs to compare using the registry key format (or canonicalize both sides), and the test should use the live registry shape.

paulbalaji (commented):
Blocking: this does not match the live Tron warp-route map shape, so it likely leaves the list labels blank.

paulbalaji (commented):
Re-reviewed `1d40c90`. No blocking findings.

I verified the new canonical comparison against the live `hyperlane-registry@main` Tron map and a real Hasura sample (`0xd29637bc06874c59c6af1a786dbca307d38d43475f482d312a5bb6c38951c05c`): sender/recipient now resolve to USDT on both sides. The focused Jest test also passes locally.

Residual note: I cannot approve my own PR via GitHub, so leaving this as a comment review.

src/utils/token.ts
```diff
@@ -1,20 +1,40 @@
...
+// recipients reaching this code have already been decoded to protocol-native
+// form (e.g. base58 for Tron). Both decode to the same 20-byte payload.
+function tryToCanonicalBytes32(address: string, protocol: ProtocolType): string | undefined {
+  try {
+    const sourceProtocol = isAddressEvm(address) ? ProtocolType.Ethereum : protocol;
+    return addressToBytes32(address, sourceProtocol).toLowerCase();
+  } catch {
+    return undefined;
+  }
+}
```
Xaroz: will this work with the `og` page?

paulbalaji: Yes, but not through this helper. The OG route is edge runtime and uses its own path in `src/pages/api/og.tsx` + `src/utils/yamlParsing.ts`.

For OG, `fetchMessageForOG()` keeps `sender` / `recipient` as raw `0x...` bytea hex, and `fetchWarpRouteMap()` stores route keys via `normalizeAddressToHex(token.addressOrDenom)`. For Tron today both sides are hex (`0xbf807...`), so the OG lookup already matches. The table/detail path is different because `postgresByteaToAddress()` decodes the message recipient to Tron base58 before lookup, which is what this PR fixes.

src/utils/token.ts
```diff
@@ -1,20 +1,40 @@
...
+import { addressToBytes32, isAddressEvm, objKeys, ProtocolType } from '@hyperlane-xyz/utils';
+
+// Normalize an address to lowercase bytes32 hex so registry keys and
+// resolved message addresses can be compared regardless of source form.
+// Tron warp routes are stored in the registry as EVM-shaped hex (`0x...`);
+// recipients reaching this code have already been decoded to protocol-native
+// form (e.g. base58 for Tron). Both decode to the same 20-byte payload.
+function tryToCanonicalBytes32(address: string, protocol: ProtocolType): string | undefined {
+  try {
+    const sourceProtocol = isAddressEvm(address) ? ProtocolType.Ethereum : protocol;
```
Xaroz, line 17: Routing any 40-hex `0x...` through the EVM decoder is the right pragmatic choice for the Tron-stored-as-hex case, but worth a one-line comment that this is shape-based, not protocol-correct. A Starknet field element whose value happens to fit in 160 bits would also match `isAddressEvm` and route through the EVM path instead of the chain's Starknet decoder. In practice the registry zero-pads Starknet to 64 hex so it shouldn't trigger, and even if it did both sides would route through the same path → still match. Just worth flagging the assumption explicitly so a future Starknet-touching change doesn't get surprised.

src/utils/token.ts
```diff
@@ -1,20 +1,40 @@
...
+
+  const normalizedAddress = tryToCanonicalBytes32(address, protocol);
+
+  for (const tokenAddress of objKeys(chain)) {
+    const normalizedKey = tryToCanonicalBytes32(tokenAddress, protocol);
+    if (normalizedAddress && normalizedKey && normalizedAddress === normalizedKey) {
+      return chain[tokenAddress];
+    }
+    // Fallback for non-address denoms (e.g. IBC denoms) that can't be canonicalized.
+    if (!normalizedKey && tokenAddress.endsWith(address)) {
```
Xaroz, line 42: Asymmetric guard: the fallback only fires when `normalizedKey` is undefined. If `normalizedAddress` is undefined while `normalizedKey` is defined (malformed lookup input vs valid registry key), neither branch matches and we silently fall through to the next iteration / final `undefined`.

Likely fine in practice — the lookup address should always be canonicalizable when the key is — but consider either:

- Symmetric guard: `if ((!normalizedKey || !normalizedAddress) && tokenAddress.endsWith(address))`, or
- An explicit comment noting that asymmetric failure is intentional and means malformed input → return `undefined`.

Makes the contract obvious to a future reader.

src/features/messages/utils.test.ts
```diff
@@ -183,6 +188,94 @@ describe('parseWarpRouteMessageDetails', () => {
...
+      const result = parseWarpRouteMessageDetails(
+        message,
+        warpRouteChainAddressMap,
+        chainMetadataResolver,
+      );
+
+      expect(result).toBeDefined();
+      expect(result!.originToken.symbol).toBe('USDT');
+      expect(result!.destinationToken.symbol).toBe('USDT');
+      expect(result!.transferRecipient).toBe(TRON_ROUTER_BASE58);
```
Xaroz, line 276: Consider a companion test for the Tron-origin → EVM-destination direction. The current test exercises `destinationToken` resolution; the symmetric case exercises `originToken` resolution against the same hex-vs-base58 mismatch, which is the same code path but a different code path through `parseWarpRouteMessageDetails`. Cheap to add and locks in both directions.

Xaroz (commented):
## Review Summary

Solid fix at the right layer — moving the comparison into a canonical bytes32 form means every caller of `getTokenFromWarpRouteChainAddressMap` benefits, and as a side effect this also kills a latent EIP-55 case-sensitivity risk in the old `endsWith` against checksummed registry keys. Keeping `endsWith` as a fallback for non-canonicalizable IBC denoms is the right call.

A couple of inline notes on the matching logic. Otherwise mostly suggestions around test coverage and one minor naming nit.

### Suggestions

- **Test coverage is thin.** Only EVM origin → Tron destination is covered. Worth adding:
  - **Tron origin → EVM destination** — mirror image, exercises `originToken` resolution against the same encoding mismatch.
  - **IBC-denom fallback regression** — the comment claims `endsWith` is preserved for non-canonicalizable keys (e.g. `factory/neutron.../eclip`), but no test pins this. A future refactor could regress it silently.
  - **Pure-EVM case mismatch** — registry checksummed vs runtime lowercased (or vice versa). Locks in the latent EIP-55 fix as intentional rather than incidental.
- **Naming nit** (subjective): `tryToCanonicalBytes32` reads slightly awkwardly. `canonicalBytes32` (with the `| undefined` return implying the try semantic) or `toCanonicalBytes32OrUndefined` would be clearer.
- **Per-call key normalization cost.** Each lookup now runs `addressToBytes32` over every registry key for the chain. Not a real concern given upstream React-layer memoization, but if profiling ever flags this, normalizing keys once at `buildWarpRouteMaps` time would eliminate it.

### Out of scope (already flagged in the PR description, just confirming)

- `useCollateralStatus.ts` and `useWarpRouteBalances.ts` still gated on `ProtocolType.Ethereum` only — Tron destination message detail pages will still miss balance/collateral panels until those are widened. Worth tracking as a separate issue so it doesn't slip.

src/utils/token.ts
paulbalaji: Added that comment and renamed the helper to `toCanonicalBytes32OrUndefined` in `4a79b9d`. The comment now calls out that 20-byte hex routing is shape-based and exists for registry entries like Tron stored as EVM-shaped hex.

src/utils/token.ts
paulbalaji: Kept the asymmetric fallback and added the explicit comment in `4a79b9d`: if the registry key is a valid address but the lookup input is malformed, we do not fall back to suffix matching. That keeps malformed lookup input from accidentally matching a valid address key.

src/features/messages/utils.test.ts
paulbalaji: Added in `4a79b9d`: `matches a Tron origin token whose registry key is EVM-shaped hex`. I also added focused tests for EVM checksum/lowercase matching and the non-address denom suffix fallback in `src/utils/token.test.ts`.

Xaroz (approved):

PR #323, opened 2026-04-27
fix(warp): use origin token decimals to decode wire amount

Xaroz:
## Summary

Removes the `wireDecimals` (max-decimals-in-route) fallback used when an origin token has no `scale` field. The wire amount is always in the origin token's native decimal space — there is no implicit max-decimals normalization to invert.

### The bug this fixes

USDT on the SVM ↔ Tron / EVM route: sending from Eclipse or Solana to Tron displayed **"USDT 0"** because:

- BSC USDT in that route has `decimals: 18, scale: {1, 1e12}` (compresses 18-dec to 6-dec wire format).
- SDK's `buildWarpRouteMaps` sets `wireDecimals = Math.max(decimals) = 18` for **every** token in the route, including the SVM ones.
- Eclipse and Solana USDT entries have **no `scale`** field in the registry, so the explorer fell into the `originToken.scale === undefined` branch of `getEffectiveDecimals` and used `wireDecimals: 18`.
- On-chain SVM (`HyperlaneToken`) sends wire amount = 1e6 for 1 USDT (`decimals=6, remote_decimals=6`).
- `fromWei("1000000", 18)` ≈ `0.000000000001` → renders as `0`.

Tron → SVM works because Tron has `scale: {1, 1}` set, hitting the early-return branch that uses origin's native decimals.

### Why `wireDecimals` is wrong in general

On-chain truth (verified in `solidity/contracts/token/libs/TokenRouter.sol` and `rust/sealevel/libraries/hyperlane-sealevel-token/src/accounts.rs`):

- **EVM `TokenRouter`** applies `messageAmount = localAmount * scaleNumerator / scaleDenominator`, with the constructor enforcing `scale > 0` — defaulting to `{1, 1}` (identity) when the deployment didn't specify scale.
- **SVM `HyperlaneToken`** stores `decimals` and `remote_decimals` and applies `convert_decimals(amount, decimals, remote_decimals)` — also identity when both are equal, which is the SVM equivalent of `scale: {1, 1}`.

There is no implicit "normalize to max(decimals) of route" anywhere in the contracts. `wireDecimals = Math.max(decimals)` was a heuristic that only ever produced the right number when the origin chain happened to be the max-decimal chain in the route (e.g. BLEND: Fluent 18-dec origin = wireDecimals 18). For any route with a scale-down chain (USDT, where actual wire format is 6-dec, not 18), it gave wrong results.

The SDK's transfer-side helper (`messageAmountFromLocal`) already defaults `scale: undefined → {1, 1}` (`hyperlane-monorepo/typescript/sdk/src/utils/decimals.ts`), so this PR aligns the explorer's display path with the SDK's transfer path.

### What changes

- `getEffectiveDecimals` and `EffectiveDecimalsToken` deleted from `src/utils/warpRouteAmounts.ts`. Callers (`src/features/messages/utils.ts`, `src/pages/api/og.tsx`) read `originToken.decimals` directly.
- `src/features/messages/warpFees/fetchWarpFees.ts`: `sentAmountToLocal` returns the wire amount unchanged when `scale === undefined` (no more `convertDecimalsToIntegerString(wireDecimals, native, ...)` fallback). Param `nativeDecimals` dropped — was a no-op.
- `src/utils/yamlParsing.ts`: removed dead `WarpToken.wireDecimals` field, the `Math.max(decimals)` reducer, and the assignment.
- `src/types.ts`: `WarpRouteDetails` now uses plain `TokenArgs` instead of the SDK's `TokenArgsWithWireDecimals` alias — no consumer reads `wireDecimals`, so the field shouldn't leak through our public type.
- `src/consts/tokenStandards.ts` deleted (the `COSMOS_STANDARDS` set was only used by the removed Cosmos branch in `getEffectiveDecimals`).
- Tests updated.

### Drive-by

Adds `/commit` and `/resolve-pr-reviews` Claude skills to `.claude/skills/`, adapted from the `hyperlane-warp-ui-template` versions to match this repo's pnpm scripts and conventions.

### Follow-up (not in this PR)

A matching change in `hyperlane-warp-ui-template` (`formatMessageAmount` in `scaleUtils.ts`) is still needed — its sidebar TransactionHistory has the same bug, masked by localStorage cache.

## Test plan

- [x] `pnpm run typecheck` — clean
- [x] `pnpm run lint` — 0 warnings, 0 errors
- [x] `pnpm run format`
- [x] `pnpm exec jest --testPathIgnorePatterns="scheduleWhenIdle|fetchPiChainMessages"` — 9/9 suites, 81/81 tests pass (the two excluded suites are pre-existing failures unrelated to this PR: missing `jest-environment-jsdom` and a stale Sepolia message fixture)
- [ ] Smoke-test SVM → Tron USDT message in dev — should now show the correct amount instead of "0"
- [ ] Smoke-test Tron → SVM USDT message in dev — should still display correctly (regression check)
- [ ] Smoke-test BLEND Fluent ↔ Solana — wire format = 18 dec, both directions should still display correctly


Xaroz:
@claude review

paulbalaji (approved):
No blocking findings. Reviewed the existing Claude and CodeRabbit reviews and rechecked the diff/callers. The behavior change matches the intended on-chain identity-scale semantics, and CI is green.

Prior comments worth optional follow-up: stale `wireDecimals` computation/comment in `yamlParsing`, stale OG comment outside the diff, and unused `nativeDecimals` param in `sentAmountToLocal`. None are merge blockers for this fix.

Xaroz:
@claude bot: addressed all 4 findings in c6b1023.

- **Finding 1** yamlParsing dead code → removed field, reducer, and assignment. `WarpToken.wireDecimals` was a local interface field; SDK's `TokenArgsWithWireDecimals` still requires `wireDecimals: number` so test mocks of `WarpRouteChainAddressMap` keep it (but simplified the helper plumbing). Also bonus: switched `WarpRouteDetails` to plain `TokenArgs` so the SDK's wireDecimals-flavored type doesn't leak through our public types.
- **Finding 2** `sentAmountToLocal` `nativeDecimals` param → dropped. While at it, dropped the now-unused `decimals` arg to `getWarpRouteAmountParts` since it's optional and unused for the amount math.
- **Finding 3** SVM edge case where `decimals != remote_decimals` and no `scale` in registry → noted; same constraint as the SDK transfer path. Followup PR for warp-ui will land separately.
- **Finding 4** stale comment in utils.ts → removed.

Xaroz:
@coderabbitai: stale doc bullet at og.tsx:78-86 fixed in c6b1023. Replaced the 3-bullet block with a one-liner: "Wire amount = localAmount * scale; scale defaults to identity when unset, so the wire amount is in the origin token's native-decimal space."

Xaroz:
@paulbalaji: thanks for the approval. Cleaned up all three optional follow-ups in c6b1023 — yamlParsing dead code, og.tsx stale comment, and the unused `nativeDecimals` param.

PR #325, opened 2026-04-29
feat(messages): add warp route ISM details card

Xaroz:
## Summary

Adds an expandable **Warp Route Security** card on the message detail page that reads the configured ISM tree and contract owner for the warp route's origin and destination tokens — directly from on-chain contracts. Sits beneath `Warp Transfer Details`, gated by the same registry-match check, lazy-loaded so no RPC traffic fires until the user expands.

## What it shows

- **Owner**: token contract's `owner()` address. If the owner is a Gnosis Safe (detected via a two-method probe — `getThreshold()` + `nonce()` — matching the SDK's own pattern), it gets a `Safe N/M` badge with threshold/owner-count when available. EOA owners just show the address — no badge. Owner = `0x0` shows "Ownership renounced".
- **ISM tree**: recursive view of the configured ISM, type-labelled per node (`Aggregation ISM`, `Pausable ISM`, `Trusted Relayer ISM`, `Fallback Routing ISM`, `Message ID Multisig ISM`, etc.) with each address copyable + linked to the chain's block explorer.
- **Mailbox-default fallback**: when the token's ISM is `0x0`, falls back to `Mailbox.defaultIsm()` and tags the tree root with a `mailbox default` badge.
- **Routing children labelled by chain display name** (`Arbitrum`, `Ethereum`) — looked up from chain metadata via `getChainDisplayName`, falls back to title-cased chain name.
- **Per-branch errors annotated inline** (`(Timed out 20000ms)`, `(Cycle detected)`, `(Failed to walk child)`) — siblings render normally even if one sub-domain reverts.

## How it works

Three layers, by design:

1. **`MailboxClient__factory`** reads `interchainSecurityModule()`, `mailbox()`, `owner()` from the warp route token. Parent of all warp route token standards (HypERC20, HypNative, HypERC20Collateral, etc.).
2. **`ShallowEvmIsmReader`** subclasses the SDK's `EvmIsmReader` and overrides `deriveIsmConfig(string)` to return placeholders. The SDK gives us correct refined type labels per node (Trusted Relayer vs Pausable vs Null, Fallback vs Domain vs Incremental Routing, etc.) without the SDK's all-or-nothing recursion.
3. **Custom recursive walker** drives the tree expansion with per-branch `try/catch` and concurrency 4 — so one bad sub-ISM doesn't kill the whole tree. Visited-set guards against cycles in misconfigured routing graphs.

The explorer's `HyperlaneSmartProvider` (multi-RPC fallback + retries) is reused — wrapped with a `Proxy` (`abortableProvider.ts`) that intercepts both the public surface (`send`, `perform`, `call`, `getCode`, ...) and the SmartProvider's internal retry/fallback methods (`performWithFallback`, `wrapProviderPerform`). The proxy binds `this` to the receiver so internal `this.*` retry hops re-enter the proxy and check the abort signal on every retry attempt.

Provider wrapping + SDK MultiProvider construction happens **inside `runSideWithTimeout`**, scoped to the per-side `AbortController`. The same signal is checked by the upfront router/mailbox/Safe reads, the walker, and every recursive sub-derive — so the side timeout actually halts all in-flight work, not just the local awaits.

## Caveats

- **Per-call timeout**: each ISM-node derive has a **20s timeout**. If it fires, the node renders with its coarse `moduleType`-based label (e.g. `Aggregation ISM`, `Routing ISM`) plus a `(Timed out 20000ms)` annotation; siblings keep walking. Each call has its own `AbortController` so the timeout actually halts the underlying SDK derive — without this, a per-call timeout merely resolves our await while the SDK's retry chain keeps firing requests.
- **Side-level timeout**: 60s safety net. `Promise.race` against the side abort signal so the UI surfaces "Timed out after 60s" exactly at 60s, regardless of how slowly individual in-flight calls unwind.
- **Background RPC tail after timeout** (cosmetic, unfixable from outside ethers): ethers v5's lowest-level `fetchJson` has its own uncancellable retry loop (up to 12 attempts on HTTP 429), invisible to our abort proxy. After the user-visible timeout fires, a tail of background RPC requests can continue until those fetchJson loops exhaust on rate-limited public RPCs. UI is correct (timeout shown at the bound) and no result is consumed by the dropped query — only the network tab shows the tail. A `TODO` comment in `runSideWithTimeout` documents the path forward (`throttleLimit: 1` on connection, or custom `fetchFunc`) if this ever becomes user-impacting.
- **Public-RPC variance**: ISM derivation can fan out into many RPC reads (e.g. an aggregation containing a routing ISM with 21 sub-domains). With public RPCs that rate-limit or are flaky, individual branches may time out — the rest of the tree still renders, with annotations on the failed branches.
- **EVM-only**: non-EVM sides (Cosmos, SVM, Tron) render `ISM details not available for {protocol} chains.`
- **No deep-detail per node** by design — only ISM type + address (no validator lists, paused state, threshold values, relayer URLs). Adding those would be a follow-up.

## Test plan

- [ ] Open a registry-matched warp transfer message
- [ ] Confirm the card appears under `Warp Transfer Details`, collapsed by default
- [ ] Expand → spinner → tree renders for both origin and destination
- [ ] Verify owner: address-only for EOAs, `Safe N/M` badge for Gnosis Safes, "Ownership renounced" when owner is `0x0`
- [ ] Verify routing-ISM children show chain display names (e.g. `Arbitrum`, `Ethereum`)
- [ ] Verify mailbox-default route: tree root shows `[mailbox default]` badge
- [ ] Verify per-call abort: navigate away mid-fetch → no further RPC calls in network tab (note: pre-existing fetchJson retry loops at the lowest level may continue briefly — see caveat)
- [ ] Verify timeout annotation: a flaky route shows `(Timed out 20000ms)` on the slow branch, siblings render
- [ ] Test on a deep tree (Aggregation → Routing with many domains) — confirm partial render works
- [ ] Verify a11y: tabbing through the header reaches the toggle button (announced as expanded/collapsed); clicking the `?` tooltip does not toggle the card

## Screenshots
image
image
image


Xaroz:
@claude review

src/features/messages/queries/abortableProvider.ts
```diff
@@ -0,0 +1,49 @@
...
+        ABORT_INTERCEPT_METHODS.has(prop) &&
+        typeof value === 'function'
+      ) {
+        return async (...args: unknown[]) => {
+          if (signal?.aborted) throw new AbortError();
+          return value.apply(target, args);
+        };
+      }
+      return typeof value === 'function' ? value.bind(target) : value;
+    },
```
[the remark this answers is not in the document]
Xaroz: Fixed in dff37da — switched to `value.apply(receiver, args)`. Same fix applied to the bind fall-through. SmartProvider's internal `this.performWithFallback` / `this.wrapProviderPerform` hops now re-enter the proxy on every retry.

src/features/messages/queries/fetchWarpRouteIsm.ts
```diff
@@ -0,0 +1,285 @@
...
+    wrappedProviders[chain] = wrapWithAbort(provider, input.signal);
+  }
+  const sdkMultiProvider = new MultiProvider(input.chainMetadata, {
+    providers: wrappedProviders,
+  });
+
+  const [origin, destination] = await Promise.all([
+    runSideWithTimeout(input.chainMetadata, wrappedProviders, sdkMultiProvider, input.origin, input.signal),
+    runSideWithTimeout(input.chainMetadata, wrappedProviders, sdkMultiProvider, input.destination, input.signal),
+  ]);
```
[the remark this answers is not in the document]
Xaroz: Fixed in dff37da — moved provider wrapping + SDK MultiProvider construction INSIDE `runSideWithTimeout`, scoped to the per-side `ctrl.signal`. The upfront router/mailbox/Safe reads AND the walker now all see the side-scoped abort signal. Combined with the abortableProvider receiver fix above, side timeouts actually halt the underlying RPC fan-out.

src/features/messages/queries/walkIsm.ts
```diff
@@ -0,0 +1,288 @@
...
+      const ism = new ethers.Contract(address, ISM_MODULE_TYPE_ABI, callProvider);
+      const moduleType: number = await ism.moduleType();
+      coarseLabel = MODULE_TYPE_FALLBACK_LABELS[moduleType] ?? `Type ${moduleType}`;
+    } catch (e) {
+      if (e instanceof IsmWalkAbortError) throw e;
+      return {
+        address,
+        typeLabel: 'Unknown',
+        error: e instanceof Error ? e.message : String(e),
+      };
```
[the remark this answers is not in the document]
Xaroz: Fixed in dff37da — coarse-label probe failure no longer kills the branch. Stash the error in `coarseError` and still try `deriveIsmConfigFromAddress`; only fall back to the coarse error if that also fails.

src/features/messages/queries/walkIsm.ts
```diff
@@ -0,0 +1,288 @@
...
+    try {
+      cfg = await reader.deriveIsmConfigFromAddress(address);
+    } catch (e) {
+      if (e instanceof IsmWalkAbortError) throw e;
+      // Distinguish per-call timeout from other failures
+      if (callCtrl.signal.aborted && !ctx.signal?.aborted) {
+        deriveError = `Timed out (${PER_NODE_TIMEOUT_MS}ms)`;
+      } else {
+        deriveError = e instanceof Error ? e.message : String(e);
+      }
```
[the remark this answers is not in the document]
Xaroz: Fixed in dff37da — both catches now rethrow `AbortError` (from wrapWithAbort) in addition to `IsmWalkAbortError`. Real parent aborts / side timeouts now propagate cleanly instead of being folded into per-node errors.

src/features/messages/cards/ismRender/AddressInline.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <span className="inline-flex items-center gap-1.5">
+      <span className="font-mono text-sm text-gray-700">{display}</span>
+      {!isZero && (
+        <CopyButton copyValue={address} width={12} height={12} className="opacity-60" />
+      )}
+      {link && (
+        <a target="_blank" rel="noopener noreferrer" href={link}>
+          <BoxArrowIcon width={12} height={12} />
+        </a>
```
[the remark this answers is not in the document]
Xaroz: Fixed in dff37da — added `aria-label="View on block explorer"` to the icon-only link.

src/features/messages/cards/WarpRouteIsmDetailsCard.tsx
```diff
@@ -0,0 +1,157 @@
...
+          />
+        </button>
+        <button onClick={() => setIsExpanded((v) => !v)}>
+          <ChevronIcon
+            width={20}
+            height={20}
+            direction={isExpanded ? 'n' : 's'}
+            className="text-gray-400"
+          />
+        </button>
```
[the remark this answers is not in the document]
Xaroz: Fixed in dff37da — Tooltip moved out of the toggle button (clicking the question mark no longer toggles the card), both toggle buttons get `aria-expanded` + `aria-label` so screen readers announce the expand/collapse state.

src/features/messages/queries/fetchWarpRouteIsm.test.ts
```diff
@@ -0,0 +1,301 @@
...
+      providers,
+      origin: { chainName: 'ethereum', tokenAddress: TOKEN_ADDRESS },
+      destination: { chainName: 'arbitrum', tokenAddress: TOKEN_ADDRESS },
+    });
+
+    if (result.origin.kind === 'data') {
+      expect(result.origin.value.ownerKind).toBe('eoa');
+      expect(result.origin.value.safeInfo).toBeUndefined();
+    }
+  });
```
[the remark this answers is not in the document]
Xaroz: Fixed in dff37da — added explicit `expect(result.origin.kind).toBe('data')` before each narrowing block in the three EOA / Safe owner tests.

WARP_ISM_PLAN.md
```diff
@@ -0,0 +1,155 @@
...
+         └── for each side (origin, destination):
+               runSideWithTimeout
+                 ├── compose side AbortController (parent-abort OR 60s timeout)
+                 └── fetchSide
+                       ├── MailboxClient__factory.connect(token) → reads:
+                       │     interchainSecurityModule(), mailbox(), owner()
+                       ├── if ISM is zero: Mailbox.defaultIsm() → labeled "mailbox default"
+                       ├── walkIsmTolerant → walkIsm (recursive tree walker)
+                       └── detectOwner: ISafe.getThreshold/getOwners → EOA vs Safe N/M
+```
```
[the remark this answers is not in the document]
Xaroz: Skipped — `WARP_ISM_PLAN.md` was removed from PR tracking in 24d0abb (`git rm --cached`). The file is local-only working notes, not part of the merged change.

src/features/messages/queries/fetchWarpRouteIsm.ts
```diff
@@ -0,0 +1,274 @@
...
+
+async function detectOwner(
+  provider: ethers.providers.Provider,
+  owner: string,
+): Promise<{ ownerKind: OwnerKind; safeInfo?: SafeInfo }> {
+  if (isZeroishAddress(owner)) return { ownerKind: 'eoa' };
+
+  try {
+    const safe = ISafe__factory.connect(owner, provider);
+    const [threshold] = await Promise.all([safe.getThreshold(), safe.nonce()]);
```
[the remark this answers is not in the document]
Xaroz: Addressed in dff37da — kept the call (it's the deliberate two-method Safe-detection probe matching the SDK's `EvmWarpRouteReader` pattern; a random contract might implement `getThreshold`, `nonce` cuts false positives) and added a comment explaining why.

src/features/messages/cards/WarpRouteIsmDetailsCard.tsx
```diff
@@ -0,0 +1,156 @@
...
+    originTokenAddress: warpRouteDetails?.originToken.addressOrDenom,
+    destinationChainName,
+    destinationTokenAddress: warpRouteDetails?.destinationToken.addressOrDenom,
+    enabled: isExpanded,
+  });
+
+  if (!warpRouteDetails) return null;
+
+  return (
+    <Card className={blur ? 'w-full blur-xs' : 'w-full'}>
```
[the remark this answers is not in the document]
Xaroz: Fixed in dff37da — switched to `clsx('w-full', blur && 'blur-xs')`, matches the sibling card.

src/features/messages/cards/WarpRouteIsmDetailsCard.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const [isExpanded, setIsExpanded] = useState(false);
+
+  const originChainName = useMemo(
+    () => multiProvider.tryGetChainName(message.originDomainId) ?? undefined,
+    [multiProvider, message.originDomainId],
+  );
+  const destinationChainName = useMemo(
+    () => multiProvider.tryGetChainName(message.destinationDomainId) ?? undefined,
+    [multiProvider, message.destinationDomainId],
+  );
```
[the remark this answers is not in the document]
Xaroz: Fixed in dff37da — inlined both `tryGetChainName` calls and dropped the unused `useMemo` import.

src/features/messages/queries/useWarpRouteIsm.ts
```diff
@@ -0,0 +1,93 @@
...
+          chainName: destinationChainName!,
+          tokenAddress: destinationTokenAddress!,
+        },
+        signal,
+      });
+    },
+    enabled: isReady,
+    retry: false,
+    staleTime: Infinity,
+    gcTime: 0,
```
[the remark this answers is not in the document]
Xaroz: Addressed in dff37da — combo is intentional. `staleTime: Infinity` keeps data fresh while the card is mounted (collapse/re-expand is free since the observer stays subscribed). `gcTime: 0` evicts on unmount so navigating to a different message doesn't show stale ISM from the previous one. Added a comment in the file explaining this so it doesn't get re-flagged.

src/features/messages/queries/walkIsm.ts
```diff
@@ -0,0 +1,280 @@
...
+};
+
+function labelFor(type: string): string {
+  return ISM_TYPE_LABELS[type] ?? type;
+}
+
+export async function walkIsm(ctx: WalkContext, address: string, depth = 0): Promise<IsmNode> {
+  checkAborted(ctx.signal);
+  if (depth > MAX_DEPTH) {
+    return { address, typeLabel: 'Unknown', error: 'Max recursion depth' };
```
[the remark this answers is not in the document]
Xaroz: Fixed in dff37da — added `logger.warn` with chain + address. Hitting depth 6 is unusual enough that field observability is worth the one-liner.

Xaroz:
@coderabbitai @claude — addressed all 13 review comments in dff37da. Summary:

**Critical / major (all fixed):**
- `abortableProvider.ts` — `value.apply(receiver, args)` so SmartProvider's internal `this.*` retry hops re-enter the proxy
- `fetchWarpRouteIsm.ts` — provider wrapping + SDK MultiProvider construction moved INSIDE `runSideWithTimeout`, scoped to per-side `ctrl.signal` so the side timeout reaches all reads (router/mailbox/Safe + walker)
- `walkIsm.ts:140` — flaky `moduleType()` no longer kills the branch; stash error and still try `deriveIsmConfigFromAddress`
- `walkIsm.ts:163` — propagate `AbortError` (in addition to `IsmWalkAbortError`) so real aborts cancel cleanly

**Minor (all fixed):**
- `clsx()` for conditional className
- Drop unnecessary `useMemo` around `tryGetChainName`
- Header a11y: Tooltip moved out of toggle button, `aria-expanded` + `aria-label` on both toggle buttons
- `aria-label` on icon-only block-explorer link
- Tests now `expect(...kind).toBe('data')` before narrowing
- `logger.warn` on `MAX_DEPTH` hit
- Comment explaining the `getThreshold + nonce` two-method Safe-detection probe (matches SDK pattern)
- Comment explaining the `staleTime: Infinity` + `gcTime: 0` combo (intentional — fresh while mounted, evict on unmount)

**Skipped:**
- markdownlint fence labels in `WARP_ISM_PLAN.md` — file was removed from PR tracking in 24d0abb (`git rm --cached`); local-only working notes, not part of the merged change.

Format / typecheck / lint / tests all pass.

src/features/messages/queries/fetchWarpRouteIsm.ts
```diff
@@ -0,0 +1,267 @@
...
+  const wrappedProviders: ChainMap<ethers.providers.Provider> = {};
+  for (const [chain, provider] of Object.entries(rawProviders)) {
+    wrappedProviders[chain] = wrapWithAbort(provider, ctrl.signal);
+  }
+  const sdkMultiProvider = new MultiProvider(chainMetadata, {
+    providers: wrappedProviders,
+  });
+
+  try {
+    return await fetchSide(chainMetadata, wrappedProviders, sdkMultiProvider, side, ctrl.signal);
```
paulbalaji: **Major — side timeout is soft.**

`wrapWithAbort` only short-circuits at *method-invocation* time. Once an RPC is in flight, awaiting a response, the abort doesn't cancel the underlying fetch — the await keeps blocking. So when the 60s timer fires `ctrl.abort()`, any subsequent calls short-circuit, but a side stuck in a single hung RPC (public RPCs that don't honor request cancellation) leaves `await fetchSide(...)` waiting indefinitely. Net effect: spinner pinned past the timeout, defeating the safety net.

Fix: race `fetchSide(...)` against an explicit abort/timeout promise so the side resolution itself is bounded:

```ts
const abortPromise = new Promise<never>((_, reject) =>
  ctrl.signal.addEventListener('abort', () => reject(new AbortError())),
);
return await Promise.race([
  fetchSide(chainMetadata, wrappedProviders, sdkMultiProvider, side, ctrl.signal),
  abortPromise,
]);
```

src/features/messages/queries/walkIsm.ts
```diff
@@ -0,0 +1,285 @@
...
+async function walkChildrenInParallel(
+  ctx: WalkContext,
+  entries: Array<{ label?: string; address: string }>,
+  depth: number,
+): Promise<IsmChild[]> {
+  const out: IsmChild[] = new Array(entries.length);
+  let cursor = 0;
+
+  const workers: Promise<void>[] = [];
+  for (let w = 0; w < Math.min(CHILD_CONCURRENCY, entries.length); w++) {
```
paulbalaji, line 276: **Major — no global node budget or visited-cycle guard.**

`CHILD_CONCURRENCY = 4` is per sibling group, not global, so recursion multiplies parallelism (4 × 4 × 4 …). With `MAX_DEPTH = 6` and worst-case fan-out (e.g., Aggregation-of-4 at every level), one side can attempt thousands of derive calls and the only ceiling is the 60s side timeout. There's also no visited-set, so a Routing → Aggregation → Routing cycle just bottoms out at `MAX_DEPTH`.

Suggest: a `Set<string>` of visited addresses passed through `WalkContext`, plus a hard global node-count cap (e.g., 64), with a clear "tree truncated" annotation when the cap is hit.

src/features/messages/queries/useWarpRouteIsm.ts
```diff
@@ -0,0 +1,97 @@
...
+  const explorerMultiProvider = useReadyMultiProvider();
+
+  const queryKey = useMemo(
+    () => [
+      'warpRouteIsm',
+      originChainName,
+      originTokenAddress,
+      destinationChainName,
+      destinationTokenAddress,
+    ],
```
paulbalaji, line 50: **Medium — query key omits `multiProviderVersion`.**

Project convention (see `useWarpRouteBalances`, `useMessageDeliveryStatus`, `useWarpFees`, `useCollateralStatus`, `useActiveRebalances`, `usePiChainMessageQuery`) is to include `multiProviderVersion` in any query key whose result depends on the provider's metadata / RPC config. Combined with `staleTime: Infinity`, the current key keeps stale ISM data after a registry / RPC config change. Add `useMultiProviderVersion()` and append it to the key + dep array.

src/features/messages/queries/walkIsm.ts
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    checkAborted(ctx.signal);
+
+    // Build a per-call MultiProvider that wraps THIS call's provider with the
+    // call-scoped abort signal. The SDK reader uses it; when the per-call
+    // timeout fires, all in-flight + pending RPCs for this call short-circuit.
+    const callMP = new MultiProvider(ctx.chainMetadata, {
+      providers: { [ctx.chainName]: callProvider },
+    });
```
paulbalaji, line 171: **Medium — `MultiProvider` rebuilt per node.**

The `MultiProvider` constructor walks the full registry through `ChainMetadataManager`. For a 200-chain registry and a 30-node ISM tree that's ~6,000 redundant initializations per side. Reuse the side-level `sdkMultiProvider` and only swap the chain's provider for the call-scoped one — or drop the per-call `MultiProvider` entirely and accept side-level abort granularity. The per-call 20s timeout is the only thing the per-call MP buys you, and the per-call wrapped `callProvider` + parent-abort listener already give you cancellation.

src/features/messages/queries/walkIsm.ts
```diff
@@ -0,0 +1,285 @@
...
+function extractChildAddresses(
+  cfg: DerivedIsmConfig,
+  chainMetadata: ChainMap<ChainMetadata>,
+): Array<{ label?: string; address: string }> {
+  const out: Array<{ label?: string; address: string }> = [];
+  const type = cfg.type as string;
+  const resolver = createChainMetadataResolver(chainMetadata);
+  const display = (chainName: string) => getChainDisplayName(resolver, chainName);
+
+  if (type === IsmType.AGGREGATION || type === IsmType.STORAGE_AGGREGATION) {
```
paulbalaji, line 230: **Medium — silently drops unknown `IsmType` variants.**

Hardcoded `if`/`else if` over a known enum set. If the SDK adds a new aggregation/routing variant, children just disappear from the rendered tree with no signal. Add a `default:` branch (or a final `else`) with `logger.warn({ type, address }, 'Unhandled IsmType, children skipped')` so the gap is visible in the field.

src/features/messages/queries/useWarpRouteIsm.ts
```diff
@@ -0,0 +1,97 @@
...
+    gcTime: 0,
+    refetchOnMount: false,
+    refetchOnWindowFocus: false,
+    refetchOnReconnect: false,
+  });
+
+  return {
+    data,
+    isLoading: isLoading && isReady,
+    error: error ? String(error) : undefined,
```
paulbalaji: **Low — double-prefixed error message.**

`String(error)` on an `Error` yields `"Error: <msg>"`, which gets rendered at `WarpRouteIsmDetailsCard.tsx:85` as `Failed to load ISM details: Error: <msg>`. Prefer:

```ts
error: error instanceof Error ? error.message : error ? String(error) : undefined,
```

src/features/messages/queries/fetchWarpRouteIsm.ts
```diff
@@ -0,0 +1,267 @@
+import { ISafe__factory, Mailbox__factory, MailboxClient__factory } from '@hyperlane-xyz/core';
+import type { ChainMap, ChainMetadata } from '@hyperlane-xyz/sdk';
+import { MultiProvider } from '@hyperlane-xyz/sdk';
+import { isZeroishAddress, ProtocolType, timeout } from '@hyperlane-xyz/utils';
+import { ethers } from 'ethers';
+
+import { AbortError, wrapWithAbort } from './abortableProvider';
+import { walkIsm, IsmWalkAbortError, type IsmNode } from './walkIsm';
+
+export type OwnerKind = 'eoa' | 'safe' | 'unknown';
```
paulbalaji: **Low — `OwnerKind = 'unknown'` is unreachable.**

`detectOwner` only ever returns `'eoa'` or `'safe'` — non-Safe contract owners (governance, non-Gnosis multisigs, timelocks) get misclassified as EOA. Either drop `'unknown'` from the union, or add a `provider.getCode(owner)` probe to distinguish contract vs EOA before the Safe check, and reserve `'unknown'` for the contract-but-not-Safe case.

src/features/messages/cards/WarpRouteIsmDetailsCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+          aria-expanded={isExpanded}
+          aria-label={`${isExpanded ? 'Collapse' : 'Expand'} Warp Route Security`}
+        >
+          <ChevronIcon
+            width={20}
+            height={20}
+            direction={isExpanded ? 'n' : 's'}
+            className="text-gray-400"
+          />
+        </button>
```
paulbalaji: **Low — toggle announced twice to screen readers.**

The label `<button>` (line 46) and this chevron `<button>` both carry the same `aria-label` and `aria-expanded`, and both call `setIsExpanded`. SR users hear "Expand Warp Route Security, button, collapsed" twice while tabbing. Either drop the chevron's a11y attributes and `aria-hidden` it (with `tabIndex={-1}` if you want the visual click target), or remove the redundant second button entirely and let the row-level toggle handle it.

src/features/messages/cards/ismRender/OwnerDisplay.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export function OwnerDisplay({ owner, ownerKind, safeInfo, chainName }: Props) {
+  const isZero = isZeroishAddress(owner);
+
+  return (
+    <div className="flex items-center gap-2 text-sm">
+      <span className="font-bold text-gray-800">Owner</span>
+      {isZero ? (
+        <span className="text-gray-500">No owner (immutable)</span>
```
paulbalaji: **Low — copy is inaccurate.**

`owner() == 0x0` means ownership renounced, not that the contract is immutable — proxy admin slots, role-based access, or other privileged paths can still exist. Suggest "Ownership renounced" or simply "No owner".

paulbalaji (commented):
## Review summary

Layered design — `MailboxClient__factory` → `ShallowEvmIsmReader` → custom recursive walker with per-branch try/catch + per-call abort plumbing — is the right shape for public-RPC reality. `dff37da` cleanly addresses nearly all prior CodeRabbit / claude[bot] feedback (Proxy `receiver` binding, per-side `ctrl.signal` wiring, `moduleType()` non-fatal, `AbortError` propagation defense, `MAX_DEPTH` `logger.warn`, header a11y split, `clsx`, `useMemo` removal, `AddressInline` aria-label, kind assertions in tests, Safe / gcTime intent comments). Nice turnaround.

A few net-new items below — none are show-stoppers, but the side-timeout-not-hard one is worth treating as a real bug because hung public RPCs can pin the spinner indefinitely past the 60s safety net.

**Major (worth resolving before merge)**
- `fetchWarpRouteIsm.ts:98` — side timeout is soft; `Promise.race` against the abort signal.
- `walkIsm.ts:256` — no global node budget or visited-cycle guard; recursion multiplies parallelism.

**Medium**
- `useWarpRouteIsm.ts:41` — queryKey omits `multiProviderVersion` (project convention).
- `walkIsm.ts:155` — `MultiProvider` rebuilt per node.
- `walkIsm.ts:215` — `extractChildAddresses` silently drops unknown `IsmType` variants.

**Low / polish**
- `useWarpRouteIsm.ts:95` — `String(error)` produces `"Failed to load ISM details: Error: <msg>"`.
- `fetchWarpRouteIsm.ts:10` — `OwnerKind = 'unknown'` is unreachable; non-Safe contract owners misclassified as EOA.
- `WarpRouteIsmDetailsCard.tsx:60` — chevron + label buttons both announce the toggle; SR users hear it twice.
- `OwnerDisplay.tsx:20` — "No owner (immutable)" is inaccurate; renounced ≠ immutable.

No additional test-coverage requests in this review, but flagging that `walkIsm.ts` and `abortableProvider.ts` are the riskiest files in the diff and currently have zero direct tests — worth a follow-up.

src/features/messages/queries/fetchWarpRouteIsm.ts
Xaroz: Fixed in 6c7f8ff — `Promise.race` against the side abort signal so the user-visible result is bounded exactly by SIDE_TIMEOUT_MS. Added a TODO documenting the fetchJson-level retry tail (uncancellable from outside ethers v5; would need `throttleLimit: 1` on connection or custom `fetchFunc` to fully kill — non-blocking for now).

src/features/messages/queries/walkIsm.ts
Xaroz: Fixed in 6c7f8ff — added a `visited: Set<string>` to `WalkContext` (one fresh Set per side). Cycle → branch labelled `(Cycle detected)`, no duplicate RPCs. Kept `MAX_DEPTH = 6` as the per-branch cap; skipped the global node budget since MAX_DEPTH × concurrency 4 already bounds total work reasonably.

src/features/messages/queries/useWarpRouteIsm.ts
Xaroz: Fixed in 6c7f8ff — added `multiProviderVersion` to both the queryKey and its deps array, matching `useWarpRouteBalances`.

src/features/messages/queries/walkIsm.ts
Xaroz: Skipped — the per-call MultiProvider construction is intentional: each `walkIsm` node has its own `AbortController` + per-call wrapped provider, and pre-populating an MP with a per-call provider is the cleanest way to scope the abort to that single derive without affecting parallel siblings. The construction cost is JS-only (no I/O); for our typical tree size the overhead is negligible. Happy to revisit if it shows up in a profile.

src/features/messages/queries/walkIsm.ts
Xaroz: Fixed in 6c7f8ff — added `logger.debug` for any `IsmType` we don't have a children handler for. Most unhandled types are genuinely childless leaves (Multisig / Pausable / TrustedRelayer / etc.) so debug-level rather than warn; if a new SDK container type lands it'll show up in console.

src/features/messages/queries/useWarpRouteIsm.ts
Xaroz: Fixed in 6c7f8ff — `error instanceof Error ? error.message : String(error)`, drops the redundant `Error:` prefix.

src/features/messages/queries/fetchWarpRouteIsm.ts
Xaroz: Fixed in 6c7f8ff — dropped `'unknown'` from `OwnerKind` (now `'eoa' | 'safe'`). Considered adding a `getCode` probe to surface non-Safe contract owners as a separate kind, but reverted because EIP-7702 (Pectra) makes EOAs return non-empty code when delegated to a smart account — would mis-classify smart-wallet EOAs as contracts. False-positive cost too high; non-Safe owners just show the address.

src/features/messages/cards/WarpRouteIsmDetailsCard.tsx
Xaroz: Fixed in 6c7f8ff — collapsed to a single toggle button wrapping LockIcon + title + chevron. Tooltip is nested inside as a `<span role="presentation">` with click/keyboard `stopPropagation` so the help icon doesn't trigger the toggle. Single ARIA announcement now.

src/features/messages/cards/ismRender/OwnerDisplay.tsx
Xaroz: Fixed in 6c7f8ff — changed to `Ownership renounced`.

Xaroz:
@paulbalaji @claude — addressed all round-2 review items in 6c7f8ff. Inline replies on each of paulbalaji's 9 anchored comments.

**Major / Medium (all fixed except one skip):**
- Side timeout via `Promise.race` against the abort signal (hard 60s user-visible bound).
- Visited-set cycle guard on `WalkContext` (kept `MAX_DEPTH` as per-branch cap; skipped global node budget since MAX_DEPTH × concurrency already bounds total work).
- `multiProviderVersion` added to `useWarpRouteIsm`'s queryKey, matching `useWarpRouteBalances`.
- `logger.debug` for unhandled `IsmType` variants in `extractChildAddresses`.
- `MultiProvider` per-node construction — **skipped intentionally**. Per-call MP is the cleanest way to scope per-call abort signals without affecting parallel siblings; construction cost is JS-only.

**Low (all fixed):**
- `error.message` instead of `String(error)` to drop the `Error:` prefix.
- Dropped `'unknown'` from `OwnerKind`. (Considered a `getCode` probe to surface non-Safe contracts — reverted because EIP-7702 delegated EOAs would mis-classify as contracts.)
- Single toggle button so the expand/collapse is announced once. Tooltip nested with `role=presentation` + `stopPropagation`.
- "Ownership renounced" instead of "No owner (immutable)".

**claude[bot] nit (fixed):**
- `walkIsmTolerant` + `walkChildrenInParallel` now rethrow `AbortError` in addition to `IsmWalkAbortError`.

**Known caveat** (also added to PR description): ethers v5's `fetchJson` has its own uncancellable retry loop (up to 12 attempts on HTTP 429) that's invisible to our abort proxy. After the user-visible timeout fires, a tail of background RPC requests can continue until those fetchJson loops exhaust on rate-limited public RPCs. UI is correct (timeout shown at the bound) and no result is consumed by the dropped query — only the network tab shows the tail. A `TODO` in `runSideWithTimeout` documents the path forward (`throttleLimit: 1` on connection / custom `fetchFunc`) if it ever becomes user-impacting.

Format / typecheck / lint / 9 tests all pass.

src/features/messages/queries/walkIsm.ts
```diff
@@ -0,0 +1,300 @@
...
+  return ISM_TYPE_LABELS[type] ?? type;
+}
+
+export async function walkIsm(ctx: WalkContext, address: string, depth = 0): Promise<IsmNode> {
+  checkAborted(ctx.signal);
+  if (depth > MAX_DEPTH) {
+    logger.warn(`ISM walk hit max depth (${MAX_DEPTH}) at ${address} on ${ctx.chainName}`);
+    return { address, typeLabel: 'Unknown', error: 'Max recursion depth' };
+  }
+  const lowerAddr = address.toLowerCase();
```
paulbalaji, line 129: **Medium — `visited` is side-global, not path-local; valid DAGs render as `(Cycle detected)`.**

`ctx.visited` is a single `Set` shared across the whole side walk, including parallel siblings. A legitimate DAG — e.g. an Aggregation whose two slots both reference the same shared `Pausable` ISM, or two Routing branches sharing a common Multisig leaf — would walk the first occurrence normally and label every subsequent occurrence `(Cycle detected)`. That's a misleading UI for a config that's structurally fine.

True cycle detection wants a path-local set (recursion-stack semantics): add the address on entry, remove on exit.

```ts
ctx.visited.add(lowerAddr);
try {
  // ... existing walk body ...
} finally {
  ctx.visited.delete(lowerAddr);
}
```

If you also want to avoid re-deriving on shared modules, layer a separate per-side dedupe cache (`Map<address, IsmNode>`) on top of the path-local cycle set — but that's optional. The minimum correct fix is path-local.

src/features/messages/queries/fetchWarpRouteIsm.ts
```diff
@@ -0,0 +1,296 @@
...
+      },
+    };
+  }
+
+  try {
+    checkAborted(signal);
+    const router = MailboxClient__factory.connect(side.tokenAddress, provider);
+
+    const [ismAddrRaw, mailboxAddr, ownerRaw] = await Promise.all([
+      timeout(router.interchainSecurityModule(), PER_CALL_TIMEOUT_MS),
```
paulbalaji, line 191: **Medium — `timeout()` is local-only; underlying RPCs keep running after the per-call timeout fires.**

`timeout(promise, ms)` from `@hyperlane-xyz/utils` is a pure promise race — it rejects after `ms` but doesn't pass an abort signal to the underlying call. So when one of the upfront router reads (`interchainSecurityModule()` / `mailbox()` / `owner()`) hangs past 20s, `Promise.all` rejects and the side returns a clean error to the user… but the wrapped provider's in-flight `fetchJson` keeps running its retry chain.

The 60s `Promise.race` from the prior round bounds the user-visible result, but it only triggers `ctrl.abort()` via the side timer. When a per-call `timeout()` completes the side faster than 60s, control returns through `runSideWithTimeout`'s `try`, the `finally` clears the timer, and `ctrl` is never aborted — wrapped provider never short-circuits. Same story for `mailbox.defaultIsm()` / Safe probes.

Two options:

```ts
// Option A: abort ctrl on any side completion (smaller change)
} finally {
  ctrl.abort();
  clearTimeout(timer);
  parentSignal?.removeEventListener('abort', parentListener);
}
```

```ts
// Option B: wrap each upfront read with its own per-call AbortController
//          (mirrors the walkIsm per-node pattern at walkIsm.ts:135)
```

Option A matches the "side is done, kill any tail" intent already present and is a 1-liner.

src/features/messages/cards/WarpRouteIsmDetailsCard.tsx
```diff
@@ -0,0 +1,164 @@
...
+            toggle the card. The tooltip itself is hover-driven so this is
+            mostly cosmetic, but it preserves the click-to-not-toggle
+            expectation a screen reader / keyboard user might have. */}
+        <span
+          role="presentation"
+          onClick={(e) => e.stopPropagation()}
+          onKeyDown={(e) => e.stopPropagation()}
+          onKeyUp={(e) => e.stopPropagation()}
+        >
+          <Tooltip
```
paulbalaji, line 62: **Low (cleanup) — partial walk-back from my prior comment.**

I flagged this more strongly than it deserves last round. `Tooltip` does render an `<a>` (`@hyperlane-xyz/widgets/dist/components/Tooltip.js:8`), but it has no `href` — which means it isn't formally "interactive content" per the HTML content model, so the nested-interactives spec violation I implied isn't exactly right. The `role="presentation"` + `stopPropagation` you added handle event leakage cleanly.

Still, structurally cleaner to move the Tooltip outside the toggle button and place it adjacent (same visual layout, no question about how AT or future Tooltip changes interact with the surrounding `<button>`). Not blocking — happy for this to land as-is.

paulbalaji (commented):
## Review summary (round 2)

`6c7f8ff` cleanly addresses 7 of 9 findings from the prior round — `Promise.race` timeout, `multiProviderVersion` in queryKey, `logger.debug` for unknown `IsmType`, `error.message` vs `String(error)`, `OwnerKind` drops `'unknown'`, "Ownership renounced" copy, and the bonus `AbortError` rethrow in `walkIsmTolerant` + `walkChildrenInParallel`. Trace-verified each.

For the two deferred items: per-node `MultiProvider` construction has a valid abort-scoping rationale (siblings run concurrently; reusing one MP would leak abort signals between them via cached provider refs in `EvmIsmReader`). Cost is ~5–20ms per construction × ~60 per side expand — not strictly "negligible" but user-imperceptible against the gated network I/O. Accepted as-is; an optional follow-up to subclass `EvmIsmReader` accepting a direct provider would sidestep the per-call MP entirely. Global node budget — accepted as follow-up; cycle guard covers the realistic misconfig class and the 60s side timeout is the practical DoS bound. Worst-case non-cyclic fanout (Aggregation-of-4 every level, MAX_DEPTH=6) is theoretical with no field evidence.

Two new items found this round (one introduced by `6c7f8ff`, one a leftover from the original soft-timeout class). Plus one cleanup nit on the Tooltip placement that I'd over-stated last round. None blocking.

With the visited-set + per-call-timeout items addressed, this is ready to approve.

paulbalaji (approved):
Approving. Two items from my last review remain open and are worth addressing as small follow-ups (non-blocking) — the path-local visited-set fix (`walkIsm.ts:125`) and aborting `ctrl` in `runSideWithTimeout`'s `finally` (`fetchWarpRouteIsm.ts:184`). Both are bounded in scope and can land separately. Solid execution across two review rounds — thanks for the careful turnaround.

src/features/messages/queries/walkIsm.ts
Xaroz: Fixed in 619c122 — switched to recursion-stack semantics. `ctx.visited.add(lowerAddr)` on entry, `ctx.visited.delete(lowerAddr)` in the `finally`. Sibling references to the same shared sub-ISM walk normally now; only true back-edges on the current path trip `(Cycle detected)`.

src/features/messages/queries/fetchWarpRouteIsm.ts
Xaroz: Fixed in 619c122 — Option A: `ctrl.abort()` in `runSideWithTimeout`'s `finally`. Any side completion path (success, per-call `timeout()` rejection, side timeout, parent abort) now abort `ctrl`, so wrapped providers short-circuit any abandoned in-flight chains. Idempotent in already-aborted cases.

src/features/messages/cards/WarpRouteIsmDetailsCard.tsx
Xaroz: Thanks for the walk-back — going to leave as-is per your "happy for this to land" call. The `role="presentation"` + `stopPropagation` handles the event leakage cleanly, and moving the Tooltip outside the button trades one structural concern (nested-not-quite-interactive) for another (visual placement that doesn't sit naturally next to the title). Easy follow-up if Tooltip's internals ever change.

PR #329, opened 2026-05-18
chore: migrate to pnpm 11

paulbalaji:
## Summary
- Bumped the repo package manager pin to pnpm 11.1.3.
- Moved pnpm overrides into `pnpm-workspace.yaml` for pnpm 11.
- Moved `enable-pre-post-scripts` out of `.npmrc` and into `pnpm-workspace.yaml`, then removed the non-auth `.npmrc`.

## Validation
- `pnpm@11.1.3 install --lockfile-only --ignore-scripts`
- `pnpm install --frozen-lockfile --ignore-scripts`
- `git diff --check`

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -2,3 +2,12 @@
 minimumReleaseAge: 10080
 minimumReleaseAgeExclude:
   - '@hyperlane-xyz/*'
+overrides:
+  tronweb>ethers: ^6.13.5
+  ethers: ^5.7
+  sha.js: 2.4.12
+  cipher-base: 1.0.5
+  elliptic: 6.6.1
+  pbkdf2: 3.1.3
+  form-data: 4.0.4
+enablePrePostScripts: true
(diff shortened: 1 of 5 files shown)


pnpm-workspace.yaml
```diff
@@ -2,3 +2,24 @@
...
   - '@hyperlane-xyz/*'
+overrides:
+  tronweb>ethers: ^6.13.5
+  ethers: ^5.7
+  sha.js: 2.4.12
+  cipher-base: 1.0.5
+  elliptic: 6.6.1
+  pbkdf2: 3.1.3
+  form-data: 4.0.4
+enablePrePostScripts: true
```
xeno097, line 13: Why do we need to explicitly enable scripts? I think the default value is true, or is just to be explicit?

paulbalaji: Yep, this is mostly to be explicit. `enablePrePostScripts` preserves the repo’s existing pre/post script behavior visibly in `pnpm-workspace.yaml` after moving pnpm config out of `package.json`. The build-script control is the `allowBuilds` block below it: those `false` entries intentionally preserve the previous “no approved dependency build scripts” posture under pnpm 11 rather than relying only on the new default behavior.

xeno097 (approved):

PR #331, opened 2026-05-18
feat: surface Sealevel + EVM cross-collateral balances in warp route graph

haggis-hyperlane:
## Summary

- Add a new Next.js API route `/api/sealevel-balance` that uses `@solana/web3.js` server-side to derive the Hyperlane escrow PDA and call `getTokenAccountBalance` against the chain's registry RPC URL.
- Wire Sealevel collateral standards (`SealevelHypCollateral`, `SealevelHypCrossCollateral`) into `useWarpRouteBalances` via a `fetch('/api/sealevel-balance?...')` call. `@solana/web3.js` stays out of the browser bundle because it's only imported from `pages/api/*`.
- Add `EvmHypCrossCollateralRouter` to the supported EVM standards (the SDK adapter factory already handles it).
- Add `EvmHypCrossCollateralRouter` and `SealevelHypCrossCollateral` to the visualization's `COLLATERAL_TOKEN_STANDARDS` so the relevant nodes render their balance/supply pill.

## Why

The Warp Route Overview graph silently dropped balance pills for the Solana node and for every EVM cross-collateral router node on routes like `CROSS/moonpay`. Reported on cross-collateral message [`0xb805…2cc3`](https://explorer.hyperlane.xyz/message/0xb805ef31a4e045e9dbb78692ead2942de12544e4d624b3117b0a41f9a3f32cc3) (USDC ETH → XO Solana).

The cause was two missing entries: the standards weren't in the visualization's collateral list, and the balance hook only enabled the EVM adapter path. Cross-collateral PRs in the SDK added the standards months ago but the explorer never picked them up.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] `pnpm run build` — `/api/sealevel-balance` registered as a dynamic server route
- [x] Pre-existing jest failures unchanged (`scheduleWhenIdle.test.ts`, `fetchPiChainMessages.test.ts`)
- [ ] Manual: load message `0xb805ef31a4e045e9dbb78692ead2942de12544e4d624b3117b0a41f9a3f32cc3`, expand Warp Route Overview, verify Solana node and EVM USDC router nodes now render a balance pill

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "@hyperlane-xyz/sdk": "33.0.2",
     "@hyperlane-xyz/utils": "33.0.2",
     "@hyperlane-xyz/widgets": "33.0.2",
+    "@solana/web3.js": "^1.98.4",
     "@tanstack/query-core": "^5.71.3",
     "@tanstack/react-query": "^5.62.3",
     "@vercel/og": "^0.8.5",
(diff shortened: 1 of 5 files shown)


Xaroz:
merged on #330

PR #332, opened 2026-05-18
feat: show cross-collateral sub-route enrollments in warp route card

haggis-hyperlane:
## Summary

- Parse each token's `connections` field (`protocol|chainName|address` refs) and `collateralAddressOrDenom` from the registry config into a typed `enrollments` list on `WarpRouteTokenVisualization`. The new field is named `enrollments` rather than `connections` to avoid colliding with the SDK's `HypTokenAdapterInput.connections` shape.
- Add `CrossCollateralEnrollments` — a new card section that lists each token in the route alongside its sub-route enrollments (chain · symbol · short address) and its collateral asset.
- Render the section under the existing graph in `WarpRouteVisualizationCard` whenever any token in the route uses a cross-collateral standard (`EvmHypCrossCollateralRouter`, `SealevelHypCrossCollateral`, `TronHypCrossCollateralRouter`).

## Why

The Warp Route Overview graph renders every token in a warp route as a flat list. For cross-collateral routers — where each token holds a distinct collateral asset and enrolls with its own subset of remote tokens — that flat list hides the sub-route structure entirely. Reported on cross-collateral message [`0xb805…2cc3`](https://explorer.hyperlane.xyz/message/0xb805ef31a4e045e9dbb78692ead2942de12544e4d624b3117b0a41f9a3f32cc3) (CROSS/moonpay, USDC ETH → XO Solana).

This uses the registry's `connections` field, which is already loaded client-side, so the new section adds no RPC calls.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run lint`
- [x] `pnpm run build`
- [x] Pre-existing jest failures unchanged
- [ ] Manual: load message `0xb805ef31a4e045e9dbb78692ead2942de12544e4d624b3117b0a41f9a3f32cc3`, expand Warp Route Overview, verify the new "Cross-collateral sub-route enrollments" section lists each token with chain, symbol, collateral asset, and enrollment chips
- [ ] Manual: load a non-cross-collateral warp route message and confirm the section does not render

### DIFF
--- a/src/features/messages/cards/WarpRouteVisualizationCard.tsx
+++ b/src/features/messages/cards/WarpRouteVisualizationCard.tsx
@@ -8,6 +8,10 @@ import { Card } from '../../../components/layout/Card';
 import HubIcon from '../../../images/icons/hub.svg';
 import { useMultiProvider } from '../../../store';
 import { Message, MessageStub, WarpRouteDetails } from '../../../types';
+import {
+  CrossCollateralEnrollments,
+  hasCrossCollateralTokens,
+} from '../warpVisualization/CrossCollateralEnrollments';
 import { useWarpRouteBalances } from '../warpVisualization/useWarpRouteBalances';
 import { useWarpRouteVisualization } from '../warpVisualization/useWarpRouteVisualization';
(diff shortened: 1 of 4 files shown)


Xaroz:
merged on #330

PR #334, opened 2026-06-05
fix(explorer): show Warped Token for CosmosIbc legs (TIA Celestia→Forma)

paulbalaji:
## Root cause

Celestia→Forma TIA transfers rendered with a blank **Warped Token** column. `parseWarpRouteMessageDetails` (`src/features/messages/utils.ts`) required **both** the origin and destination legs to resolve to a registry token by address (`getTokenFromWarpRouteChainAddressMap`, `src/utils/token.ts`).

For the celestia leg the registry token is keyed by its IBC denom `utia` (standard `CosmosIbc`, route `TIA/forma-stride-config.yaml`). But the on-chain `sender` is the Cosmos **hyperlane router app** address (bytes32 `0x726f757465725f617070…0008`, ASCII "router_app"), which has no relationship to the denom key:

- `formatAddress` returns the Cosmos address unchanged.
- `addressToBytes32("utia", Cosmos)` throws (not bech32), so the denom key can't be canonicalized.
- The `endsWith` fallback (`"utia".endsWith("0x726f…0008")`) is false.

So the origin token never resolves → both-legs requirement fails → `undefined` → blank cell. The Forma EVM leg (`0x832d26B6904BA7539248Db4D58614251FD63dC05`, `EvmNative`) matches fine.

## Change

When **exactly one** leg matches by address, derive the counterpart token via the shared warp route instead of requiring both legs to match by address:

- New `getCounterpartTokenFromWarpRoute` (`src/utils/token.ts`) walks `warpRouteIdToAddressesMap` (already built in the store by the SDK's `buildWarpRouteMaps`) from the matched leg's `{chainName, addressOrDenom}` to the token on the other chain, then looks up its metadata in `warpRouteChainAddressMap`.
- `parseWarpRouteMessageDetails` gains a `warpRouteIdToAddressesMap` parameter and invokes the counterpart lookup only when one leg is missing. The existing both-legs-match fast path is unchanged, so EVM↔EVM routes do not regress.
- Threaded the route-id map through the three call sites (`MessageTable`, `MessageDetailsInner`, transactions `MessageSummaryRow`) from the Zustand store.

## Side effects

- **Mis-attribution risk (shared collateral):** a single `addressOrDenom` can belong to multiple warp routes (reused collateral). To avoid attributing the wrong token/amount, counterpart resolution only succeeds when **exactly one** route both contains the matched leg **and** covers the other chain. The route must cover both the origin and destination chains (disambiguation by domain). If 0 or >1 candidate counterpart tokens result, we return `undefined` and render nothing — same as before. Covered by two negative unit tests (route missing the other chain; shared-collateral ambiguity).
- **Still unmatched:** transfers where **neither** leg matches by address (e.g. Cosmos→Cosmos where both legs are denom-keyed router transfers) remain blank — this PR only rescues the case where one leg is address-matchable. No behavior change for those.
- No change to fetching/indexing; messages already appear in the list, only the column was blank.

## Tests

- `src/features/messages/utils.test.ts`: new describe block for the TIA Celestia→Forma case (origin sender = router-app bytes32, dest recipient = `0x832d…dC05`) plus the two ambiguity guards. Existing tests updated for the new parameter.
- `pnpm typecheck`, `oxlint`, `oxfmt`, and the warp/token jest suites pass (12/12). Two unrelated suites (`scheduleWhenIdle`, `fetchPiChainMessages`) fail identically on `main` — pre-existing, not touched here.

### DIFF
--- a/src/features/messages/MessageDetailsInner.tsx
+++ b/src/features/messages/MessageDetailsInner.tsx
@@ -76,6 +76,7 @@ export function MessageDetailsInner({ messageId, message: messageFromUrlParams }
   const ensureWarpRouteData = useStore((s) => s.ensureWarpRouteData);
   const isWarpRouteDataLoaded = useStore((s) => s.isWarpRouteDataLoaded);
   const warpRouteChainAddressMap = useStore((s) => s.warpRouteChainAddressMap);
+  const warpRouteIdToAddressesMap = useStore((s) => s.warpRouteIdToAddressesMap);
   const [runtimeState, setRuntimeState] = useState<{
     messageId: string;
     value: MessageDetailsRuntimeState;
@@ -123,8 +124,14 @@ export function MessageDetailsInner({ messageId, message: messageFromUrlParams }
   const destinationChainName =
     chainMetadataResolver.tryGetChainName(destinationDomainId) || 'Unknown';
   const warpRouteDetails = useMemo(
-    () => parseWarpRouteMessageDetails(message, warpRouteChainAddressMap, chainMetadataResolver),
-    [chainMetadataResolver, message, warpRouteChainAddressMap],
+    () =>
+      parseWarpRouteMessageDetails(
+        message,
+        warpRouteChainAddressMap,
+        warpRouteIdToAddressesMap,
+        chainMetadataResolver,
+      ),
+    [chainMetadataResolver, message, warpRouteChainAddressMap, warpRouteIdToAddressesMap],
   );
   const txMessageCount = useTransactionMessageCount(isMessageFound ? origin.hash : undefined);
   const showTxLink = txMessageCount > 1;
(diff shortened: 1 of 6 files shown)


paulbalaji:
Superseded by the registry data fix: hyperlane-xyz/hyperlane-registry#1554.

Root cause of the blank "Warped Token" cell for `TIA/forma-stride` is that the registry modeled only the two-hop celestia→stride→forma IBC path (`utia` / CosmosIbc) and was missing the **direct** celestia→forma hyperlane leg (celestia router `0x726f…0008`, collateral `utia`). #1554 adds that leg (verified on-chain both directions).

Once #1554 merges, the explorer matches this route with **no code change** — the direct leg is keyed by the router address, so the existing both-legs-by-address logic resolves it (the explorer loads the registry at runtime). That makes this PR's matcher change unnecessary for the reported issue, and it carried a mis-attribution risk for shared-collateral addresses.

Closing in favor of the registry fix. The change here remains valid as a *standalone robustness enhancement* (resolving denom-keyed Cosmos legs that can't be matched by address) — happy to reopen and reframe it as such if we want that defense-in-depth independent of this incident.

PR #335, opened 2026-06-10
fix(messages): avoid detail prefetch on row hover

Xaroz:
## Summary
- Keep hover/focus/click prefetch for the message detail route shell and existing row stub.
- Remove the full message-detail GraphQL prefetch from table row prefetching.
- Keep top-row idle warming cheap by avoiding Hasura detail queries there too.

## Why
The previous row prefetch path fetched full message details from Hasura whenever a row was hovered. That can help only in the narrow case where a user hovers a row long enough for the detail query to finish, then clicks it. In practice, normal scrolling can move many rows under a stationary pointer and trigger hover handlers for messages the user never intended to open.

Those speculative detail queries create much more downside than upside: they burn API quota, can hit Hasura rate limits after a few scrolls, and compete with actual user-driven requests. The route shell and row stub prefetches still provide the cheap part of the optimization: navigation code is warmed and the detail page can render known row data immediately. The heavier GraphQL detail fetch now happens when the user actually navigates to a message.

## Checks
- pnpm run format
- CI=true pnpm run lint


src/features/messages/MessageTable.tsx
```diff
@@ -308,7 +303,6 @@ function LinkCell({
         className={`block h-full w-full ${aClasses || ''}`}
         onMouseEnter={onNavigateIntent}
         onFocus={onNavigateIntent}
-        onTouchStart={onNavigateIntent}
         onClick={onNavigateIntent}
```
paulbalaji, line 305: `onTouchStart` was removed from this anchor (only `onMouseEnter`/`onFocus`/`onClick` remain). This seems to work against the PR's own stated goal of keeping the *cheap* warming. The "scrolling drags rows under a stationary pointer" rationale applies to `mouseenter` hover, not to `touchstart` — a touch is a deliberate press, not incidental. On touch devices there's no hover/focus before tap, so warming now fires only on `onClick`, i.e. simultaneously with navigation, giving little head start. Since `primeDetailPage` no longer does any heavy Hasura work (just `router.prefetch` + shell chunk import + stub), consider keeping `onTouchStart` here. Also: this removal isn't mentioned in the PR description — worth calling out if intentional.

src/features/messages/MessageTable.tsx
```diff
@@ -324,18 +318,9 @@ const styles = {
   iconText: 'text-sm font-light ml-2',
 };
 
-async function prefetchMessageNavigation(
-  router: NextRouter,
-  message: MessageStub,
-  chainMetadataResolver: ChainMetadataResolver,
-  scrapedChains: ReturnType<typeof useScrapedDomains>['scrapedDomains'],
-) {
+async function prefetchMessageNavigation(router: NextRouter, message: MessageStub) {
```
paulbalaji: Nit: now that the `await prefetchMessageDetails(...)` call is gone, this function no longer `await`s anything, so `async` is no longer needed. Both call sites (`void prefetchMessageNavigation(...)` and the `Promise.all` map) work fine without it.

paulbalaji (commented):
## Review Summary

The core change is correct and well-motivated: dropping the speculative Hasura `prefetchMessageDetails` call from hover/idle/touch prefetch avoids burning API quota and hitting rate limits when normal scrolling drags many rows under a stationary pointer. The cheap warming (route `prefetch`, detail-shell chunk import, and stub cache) is preserved, and the detail page degrades gracefully — `MessageDetailsInner` still runs `useMessageQuery` because a stub never satisfies `hasCompleteMessageDetails`, so navigating to a message fetches details as before. `typecheck` passes; the relevant `prefetch.test.ts` suite passes; lint is clean.

Two things worth weighing before merge (neither is a correctness bug), plus a nit — inline below. Not blocking.

## Observations Outside This PR

This PR removes the only production caller of `prefetchMessageDetails`, leaving a now-dead subsystem behind (out of this diff, so flagging here rather than inline):

- `src/features/messages/queries/prefetch.ts:37` — `prefetchMessageDetails` now has no production callers (only `prefetch.test.ts` references it).
- `src/features/messages/queries/prefetch.ts:87` — `setPrefetchedMessageDetails` is only called from inside `prefetchMessageDetails`, so the `prefetchedMessageDetails` map is now never populated. Consequently `getPrefetchedMessageDetails()` in `MessageDetailsPage.tsx:24` always returns `undefined`, and the `prefetchedMessageDetails || message || prefetchedMessage` branch collapses to `message || prefetchedMessage`. Harmless (the real query covers it) but dead.
- The `prefetchedMessageDetailPromises` map + `prefetchedMessageVersion` snapshot/compare logic exist solely to support that dead function.
- `src/features/messages/queries/prefetch.test.ts` now exercises only dead code.

Per the repo's KISS guidance, consider a follow-up (or folding into this PR) to delete `prefetchMessageDetails`, its cache/version machinery, the always-undefined read in `MessageDetailsPage`, and the orphaned tests. Leaving it is safe, just unnecessary surface area.

_(Unrelated: `scheduleWhenIdle.test.ts` and `fetchPiChainMessages.test.ts` fail locally due to a missing `jest-environment-jsdom` dep / live-network access — pre-existing, not caused by this PR.)_

paulbalaji (approved):
Approving — the core change is correct and well-motivated, and nothing here is blocking. The points in my prior review comments (keep `onTouchStart` for cheap touch warming, the now-dead `prefetchMessageDetails` subsystem, and the `async` nit) are suggestions to address at your discretion, either here or as a follow-up.

src/features/messages/MessageTable.tsx
Xaroz: Fixed in 91a4079.

src/features/messages/MessageTable.tsx
Xaroz: Fixed in 91a4079.

Xaroz:
@coderabbitai Skipping the background prefetch ref guard. After this PR the repeated path no longer calls GraphQL, and adding stateful bookkeeping for cached route/shell/stub warming does not seem worth the extra complexity.

Xaroz:
@paulbalaji Fixed the touch handler and async nit in 91a4079. Also folded in the dead detail-prefetch cleanup by removing the unused cache path and orphaned test.

PR #336, opened 2026-06-11
feat(messages): warp fee for same-chain CCR swaps + chain-scoped route filter

haggis-hyperlane:
## Summary
- **Populate the "Warp fee" line for same-chain CCR swaps.** These swaps call `handle()` directly with no Mailbox dispatch, so they have no `DispatchId`/`SentTransferRemote`. `fetchWarpFees` relied on `DispatchId`-based log slicing and returned `null` for them, leaving the fee blank. Added a synthetic-message branch (detected via `origin == destination` + `0x00000000…` msgId prefix, matching the scraper's `syntheticCcrSwapMessageId`) that reads the swap amount from the message body (the on-chain `TokenMessage`) and scans the full receipt. Gated to txs with exactly one same-chain swap and no cross-chain dispatch so the user-pull sum stays unambiguous. Fee math is otherwise identical to the existing ERC20 path (tokens pulled − swap amount).
- **Show the warp fee rate in bps** in parentheses next to the fee amount (e.g. `0.0000015 USDT (2 bps)`), computed against the sent amount to match the on-chain `fee = amount * maxFee / (2 * halfAmount)` convention.
- **Stop other warp routes leaking into a route's search results.** The DB filter matches on address bytes alone (sender OR recipient), so a route whose token shares an address with another route on a different chain leaked in (e.g. BLEND messages under CROSS/moonpay). Added client-side filtering in `useMessageSearchQuery` that requires the matched address to be on its expected chain, resolving each `{chainName, address}` to a domain and comparing with protocol-aware `eqAddress`.

## Notes
- Client-side filter runs after the DB `LIMIT` (consistent with the existing pending-status filter). This can shrink a page below the requested size; chosen over per-chain DB clauses, which bloat the split query.
- No `@hyperlane-xyz/utils` bump needed — synthetic-id detection is structural, since the installed v33.0.2 predates the `syntheticCcrSwapMessageId` export.

## Review remediation
- `computeFeeBps` stays entirely in BigNumber — the previous `.toNumber()` threw an ethers `NUMERIC_FAULT` for large fee/amount ratios, which would have failed the whole `useWarpFees` query.
- Unresolved warp-route chains are now logged and dropped from the client-side filter rather than silently widening it (which would skip filtering and re-leak unrelated messages).
- `messageMatchesWarpRoute` exported and covered by a dedicated test suite.

## Test plan
- [x] `pnpm typecheck` — clean
- [x] `pnpm lint` — 0 warnings/errors
- [x] `pnpm test -- fetchWarpFees` — 36 passing (incl. `computeFeeBps` whole/fractional/zero/large-ratio cases)
- [x] `pnpm test -- useMessageQuery` — 6 passing (`messageMatchesWarpRoute`: origin/destination match, wrong-chain exclusion, case-insensitive, no-match, empty list)
- [ ] Preview build: verify a same-chain CCR swap message shows a "Warp fee", and that filtering by a CROSS/moonpay route no longer lists BLEND messages

### DIFF
--- a/src/features/messages/queries/useMessageQuery.ts
+++ b/src/features/messages/queries/useMessageQuery.ts
@@ -1,8 +1,9 @@
+import { eqAddress } from '@hyperlane-xyz/utils';
 import { useCallback, useMemo } from 'react';
 import { useQuery } from 'urql';
 
 import { useChainMetadataResolver } from '../../../metadataStore';
-import { MessageStatus, MessageStatusFilter } from '../../../types';
+import { MessageStatus, MessageStatusFilter, MessageStub } from '../../../types';
 import { useVisibleInterval } from '../../../utils/useVisibleInterval';
 import { useScrapedChains, useScrapedDomains } from '../../chains/queries/useScrapedChains';
 import { MessageIdentifierType, buildMessageQuery, buildMessageSearchQuery } from './build';
@@ -22,6 +23,20 @@ const PENDING_FILTER_BATCH_SIZE = 500;
 export function isValidSearchQuery(input: string) {
   if (!input) return false;
   return !!searchValueToPostgresBytea(input);
+}
+
+// A message belongs to the warp route if it was sent from the route's token on
+// the origin chain or received by it on the destination chain. eqAddress is
+// protocol-aware so this stays correct across EVM/Sealevel/Cosmos/etc.
+function messageMatchesWarpRoute(
+  message: MessageStub,
+  warpRouteDomainAddresses: Array<{ domainId: number; address: string }>,
+): boolean {
+  return warpRouteDomainAddresses.some(
+    ({ domainId, address }) =>
+      (message.originDomainId === domainId && eqAddress(message.sender, address)) ||
+      (message.destinationDomainId === domainId && eqAddress(message.recipient, address)),
+  );
 }
 
 export function useMessageSearchQuery(
@@ -56,6 +71,23 @@ export function useMessageSearchQuery(
   const isValidDestination = !destinationChainNameFilter || destDomainId !== null;
 
   const warpAddresses = warpRouteAddresses.map((a) => a.address);
+
+  // Resolve each warp route token to its domain for client-side filtering.
+  // The DB filter matches on address bytes alone (sender OR recipient), so a
+  // route whose address coincides with another route's address on a different
+  // chain leaks in (e.g. BLEND messages showing up under CROSS/moonpay).
+  // Addresses are not guaranteed unique across chains, so we additionally
+  // require the matched address to be on its expected chain.
+  const warpRouteDomainAddresses = useMemo(
+    () =>
+      warpRouteAddresses
+        .map(({ chainName, address }) => ({
+          domainId: chainMetadataResolver.tryGetDomainId(chainName),
+          address,
+        }))
+        .filter((entry): entry is { domainId: number; address: string } => entry.domainId !== null),
+    [warpRouteAddresses, chainMetadataResolver],
+  );
 
   // For pending filter, we use client-side filtering because the DB query for
   // is_delivered=false is slow (no index on absence of delivered_message record).
@@ -95,13 +127,20 @@ export function useMessageSearchQuery(
     [chainMetadataResolver, scrapedChains, data],
   );
 
-  // Apply client-side pending filter if needed
+  // Apply client-side filters. Note: these run after the DB LIMIT, so they
+  // can shrink a page below the requested size — acceptable here since the
+  // alternative (per-chain DB clauses) bloats the query and the pending
+  // filter already relies on client-side narrowing.
   const messageList = useMemo(() => {
+    let list = unfilteredMessageList;
     if (isPendingFilter) {
-      return unfilteredMessageList.filter((m) => m.status === MessageStatus.Pending);
+      list = list.filter((m) => m.status === MessageStatus.Pending);
     }
-    return unfilteredMessageList;
-  }, [unfilteredMessageList, isPendingFilter]);
+    if (warpRouteDomainAddresses.length > 0) {
+      list = list.filter((m) => messageMatchesWarpRoute(m, warpRouteDomainAddresses));
+    }
+    return list;
+  }, [unfilteredMessageList, isPendingFilter, warpRouteDomainAddresses]);
 
   const isMessagesFound = messageList.length > 0;
 
--- a/src/features/messages/queries/useMessageQuery.ts
+++ b/src/features/messages/queries/useMessageQuery.ts
@@ -4,6 +4,7 @@ import { useQuery } from 'urql';
 
 import { useChainMetadataResolver } from '../../../metadataStore';
 import { MessageStatus, MessageStatusFilter, MessageStub } from '../../../types';
+import { logger } from '../../../utils/logger';
 import { useVisibleInterval } from '../../../utils/useVisibleInterval';
 import { useScrapedChains, useScrapedDomains } from '../../chains/queries/useScrapedChains';
 import { MessageIdentifierType, buildMessageQuery, buildMessageSearchQuery } from './build';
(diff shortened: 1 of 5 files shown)


src/features/messages/warpFees/fetchWarpFees.ts
```diff
@@ -114,11 +142,27 @@ export async function fetchWarpFees(
...
+ * Fee as a rate of the sent amount, in basis points. Matches the on-chain
+ * convention (`fee = amount * maxFee / (2 * halfAmount)`), so the denominator
+ * is the amount the user sent, not the gross amount they paid. Kept to 2
+ * decimal places; returns `undefined` when the amount is zero or the rate
+ * rounds below 0.01 bps.
+ */
+export function computeFeeBps(fee: BigNumber, sentAmount: BigNumber): string | undefined {
+  if (sentAmount.isZero()) return undefined;
+  // bps × 100 = fee / amount * 10_000 * 100, using integer math for precision.
+  const bpsHundredths = fee.mul(1_000_000).div(sentAmount).toNumber();
```
paulbalaji: **Major (mechanics) / Minor (likelihood): unguarded `.toNumber()` can throw.** `BigNumber.toNumber()` raises ethers `NUMERIC_FAULT` ("overflow") once the value exceeds `Number.MAX_SAFE_INTEGER`. Here `bpsHundredths = fee.mul(1_000_000).div(sentAmount)` exceeds 2^53 when `fee/sentAmount` is large -- reachable when `sentAmount` parsed from the body is small relative to the pulled amount. The call site at line 145 is not wrapped in try/catch, so this throws out of the whole `useWarpFees` query, not just the bps suffix. (Confirmed reproducible.)

Keep it in BigNumber land and format the string from integer math, e.g.:
```ts
const bpsHundredths = fee.mul(1_000_000).div(sentAmount);
if (bpsHundredths.isZero()) return undefined;
const whole = bpsHundredths.div(100).toString();
const frac = bpsHundredths.mod(100).toString().padStart(2, '0').replace(/0+$/, '');
return frac ? `${whole}.${frac}` : whole;
```
Matches the open CodeRabbit thread on this line.

paulbalaji (commented):
## Review Summary

Re-reviewed at head `b777eb6`. Two changes: synthetic same-chain CCR warp-fee + bps display, and chain-scoped warp-route search filter. Build gates are green locally (`typecheck` clean, `lint` 0/0, `pnpm test -- fetchWarpFees` -> **34 passing**). The design is sound on both fronts -- `messageMatchesWarpRoute`'s sender<->origin / recipient<->destination pairing is correct, `eqAddress` is protocol-aware and non-throwing, and the synthetic-CCR path is properly gated (exactly one `ReceivedTransferRemote`, no `SentTransferRemote`) at the call site.

Not blocker-heavy, but two real (if low-probability) correctness findings are worth addressing before merge -- both overlap the open CodeRabbit threads, which remain unresolved with no author reply:

1. **`computeFeeBps` `.toNumber()` can throw** -- reproduced as an ethers `NUMERIC_FAULT` overflow for a large fee/sentAmount ratio (reachable when `sentAmount` is small). The call site is unguarded, so it fails the whole `useWarpFees` query, not just the bps suffix. CodeRabbit's mechanics are right; its "major" severity is overstated.
2. **Unresolved route domains fail open** -- if every route chain's domain fails to resolve, the `length > 0` guard skips filtering and falls back to address-bytes-only DB matching (i.e. the leak this PR fixes). Realistically only a transient warp-loads-before-metadata race that self-corrects, but per the repo's "fail loud, no silent fallbacks" rule it should at least warn.

Minor: post-LIMIT client filtering can yield thin pages with no backfill, and the new route-scoping helper has no tests. Details inline.

## Doc-count nit

PR description says "29 passing (7 new)"; the branch actually has 34 (12 new) -- stale figures, reality is better-covered. Worth syncing the description.

src/features/messages/queries/useMessageQuery.ts
```diff
@@ -57,6 +72,23 @@ export function useMessageSearchQuery(
...
+  // Addresses are not guaranteed unique across chains, so we additionally
+  // require the matched address to be on its expected chain.
+  const warpRouteDomainAddresses = useMemo(
+    () =>
+      warpRouteAddresses
+        .map(({ chainName, address }) => ({
+          domainId: chainMetadataResolver.tryGetDomainId(chainName),
+          address,
+        }))
+        .filter((entry): entry is { domainId: number; address: string } => entry.domainId !== null),
```
paulbalaji: **Minor: unresolved route domains fail open.** Entries whose `tryGetDomainId` returns `null` are dropped here, and the downstream guard at line 139 only filters when `warpRouteDomainAddresses.length > 0`. If *all* route chains are unresolved, chain-scoping is skipped entirely and results fall back to the address-bytes-only DB filter (`build.ts:183`) -- i.e. the cross-chain address leak this PR exists to fix.

In practice warp chains and chain metadata come from the same registry, so this is realistically only a transient warp-loads-before-metadata race that self-corrects on the next render (the `useMemo` dep on `chainMetadataResolver` recomputes). But per CLAUDE.md ("NEVER add silent fallbacks for unexpected issues -- fail loud"), the all-dropped case shouldn't silently broaden results: at minimum `logger.warn` when an entry is dropped, so a genuine registry inconsistency is visible rather than masked. Matches the open CodeRabbit thread.

src/features/messages/queries/useMessageQuery.ts
```diff
@@ -95,13 +127,20 @@ export function useMessageSearchQuery(
...
+  // alternative (per-chain DB clauses) bloats the query and the pending
+  // filter already relies on client-side narrowing.
   const messageList = useMemo(() => {
+    let list = unfilteredMessageList;
     if (isPendingFilter) {
-      return unfilteredMessageList.filter((m) => m.status === MessageStatus.Pending);
+      list = list.filter((m) => m.status === MessageStatus.Pending);
+    }
+    if (warpRouteDomainAddresses.length > 0) {
+      list = list.filter((m) => messageMatchesWarpRoute(m, warpRouteDomainAddresses));
```
paulbalaji, line 152: **Minor (by-design, flagging):** this client filter runs after the DB `LIMIT`, so a high-collision route can render a page well under `SEARCH_QUERY_LIMIT` with no "load more" / backfill. Consistent with the pending-status filter precedent -- but note pending compensates by over-fetching (`PENDING_FILTER_BATCH_SIZE` = 500), whereas the warp path keeps the base limit. Acceptable under KISS; if route-address collisions turn out common in practice, consider bumping the limit for warp searches the way pending does.

src/features/messages/queries/useMessageQuery.ts
```diff
@@ -24,6 +25,20 @@ export function isValidSearchQuery(input: string) {
   return !!searchValueToPostgresBytea(input);
 }
 
+// A message belongs to the warp route if it was sent from the route's token on
+// the origin chain or received by it on the destination chain. eqAddress is
+// protocol-aware so this stays correct across EVM/Sealevel/Cosmos/etc.
+function messageMatchesWarpRoute(
```
paulbalaji: **Minor: no tests for the new route-scoping helper.** Per CLAUDE.md ("new utility functions need `.test.ts` files"), `messageMatchesWarpRoute` is pure and easily covered. Worth a small suite: same address on a different chain -> excluded, sender<->origin match, recipient<->destination match, cross-protocol address -> `false`, and the all-unresolved-domains behavior.

paulbalaji (commented):
## Re-review @ `3a373b3`

The latest commit resolves every actionable finding from the prior review:

- **`computeFeeBps` overflow** ✅ — now stays in BigNumber and hand-formats the fixed-point string; `.toNumber()` removed. New tests cover the `1e18`/`1e21` overflow case (→ `10`, no throw) and trailing-zero trim (→ `2.5`). Spot-checked: `padStart(2,'0')` before the trim correctly preserves leading fractional zeros (e.g. `0.02`, `2.05`).
- **Unresolved domains fail open** ✅ — drop path now `logger.warn`s, so the fallback is loud instead of silent (satisfies the repo's fail-loud rule).
- **Route helper untested** ✅ — new `useMessageQuery.test.ts` covers sender↔origin, recipient↔destination, case-insensitivity, wrong-chain exclusion, no-match, and empty list.

Gates green locally: `typecheck` clean, `lint` 0/0, **42 tests pass**.

Only nit left is cosmetic — the PR description still says "29 passing (7 new)"; it's now 42. The post-LIMIT thin-page behavior remains by-design as previously noted. Approving.

paulbalaji (approved):
Approving — all actionable findings from the prior review are resolved in `3a373b3` (BigNumber-safe bps, fail-loud on unresolved domains, route-helper tests). Gates green: typecheck clean, lint 0/0, 42 tests pass.

PR #342, opened 2026-08-10
fix: link Aleo warp tokens to program pages

troykessler:
## Summary
- route Aleo warp token links to Provable program pages
- strip the account suffix from Aleo program/address values before building explorer URLs
- add URL utility coverage for Aleo program links

## Testing
- pnpm test src/utils/url.test.ts
- pnpm typecheck


paulbalaji (approved):
Verified the exact registry Aleo token shape and the live Provable program route. The protocol-gated normalization preserves bare Aleo account links, focused tests and typecheck pass locally, and exact-head CI is green. No findings.

PR #339, opened 2026-08-04
chore: upgrade dependencies and Node 24

paulbalaji:
## Summary

- upgrade all direct dependencies to current compatible versions under the release-age policy
- upgrade Hyperlane packages to core 12, registry 25.3, and SDK/utils/widgets 40
- use Vercel-supported Node 24 with pnpm 11.20, TypeScript 7, oxlint 1.77, oxfmt 0.62, Jest 30, and Tailwind CSS 4
- upgrade checkout, setup-node, cache, and pnpm setup actions to current majors
- migrate affected Hyperlane, urql, Tailwind, Next.js, lint, formatter, and test integrations
- replace the retired-chain PI live test with deterministic provider fixtures
- apply the resulting Tailwind class-order normalization

## Impact

The Explorer builds and tests on Node 24 and TypeScript 7. Tailwind 4 continues loading the existing JavaScript theme.

Ethers remains on 5.8.0 because Hyperlane SDK/widgets 40 still use ethers v5 provider contracts. Node remains on 24 because Vercel currently supports 20.x, 22.x, and 24.x.

## Observed CI timing

Adjacent fresh GitHub-hosted runs, comparing [TS6 on main](https://github.com/hyperlane-xyz/hyperlane-explorer/actions/runs/30945409865) with the [TS7 PR head](https://github.com/hyperlane-xyz/hyperlane-explorer/actions/runs/30943946129):

| Step | TS6 | TS7 | Result |
| --- | ---: | ---: | --- |
| Typecheck | 8s | 1s | 88% faster |
| Production build | 49s | 54s | Inconclusive (+5s) |

Both builds started from fresh checkouts without a persisted `.next` cache. The build delta is a single-sample comparison across the full dependency upgrade and may reflect runner or dependency variance, so it is not attributed to TypeScript 7.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm run format`
- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run test` — 14 suites, 128 tests
- `pnpm run build`
- `git diff --check`
- exact-head GitHub build, format, lint, and typecheck passed
- exact-head Socket checks passed
- exact-head Vercel preview deployed successfully
- verified custom Tailwind classes in generated production CSS

PR #343, opened 2026-08-10
fix: restore Explorer after dependency upgrade

paulbalaji:
## What

- split the unstable Zustand selector into stable scalar selectors
- force the upgraded urql client to keep GraphQL queries on POST
- retain Tailwind 3.4.19 while keeping the rest of the dependency/TypeScript upgrade
- restore hydration, data loading, and production styling after #339

### Zustand 5 render loop

Zustand 5 uses `useSyncExternalStore` and requires selector outputs to be referentially stable. `useQueryParamChainConfigSync` returned a new object on every snapshot, causing `ChainConfigSyncEffect` to enter React's maximum-update-depth error boundary during hydration.

This was the only selector in `src` returning a fresh object or array; all other selectors return existing state references or scalars.

### urql 5 transport change

urql 5 defaults queries to URL-encoded GET requests when they fit within its URL limit. Cloudflare rejects those GraphQL GETs with urql's `Accept` header (HTTP 403 without CORS headers), so the browser could not load either scraped domains or Latest Messages. The equivalent POST succeeds and returns the required CORS headers. `preferGetMethod: false` restores the previous transport explicitly.

### Tailwind 4 / widget stylesheet incompatibility

`@hyperlane-xyz/widgets/styles.css` is a precompiled Tailwind 3 bundle containing both a global preflight and prefixed `htw-*` component utilities. Tailwind 4 places the Explorer utilities in native cascade layers while the widget bundle remains unlayered, allowing its preflight to override app images, buttons, inputs, selectors, and other elements.

Putting the entire widget bundle in a lower cascade layer fixed the homepage utilities but also lowered the widget's own utilities, breaking the Origin/Destination picker and message timeline. Tailwind 4 therefore cannot be adopted safely here without first changing how the shared widget CSS is built. This PR restores the latest Tailwind 3 release (`3.4.19`) and its PostCSS pipeline; every other dependency upgrade remains.

## Verification

- `pnpm build`
- `pnpm typecheck`
- `pnpm lint`
- local production-browser comparison against `https://explorer.hyperlane.xyz`:
  - app hydrates without uncaught errors
  - Latest Messages renders from POST GraphQL requests
  - homepage header, footer, search, filters, table, and spacing match production
  - Origin/Destination picker layout, search, controls, and chain rows match production
  - supplied message timeline restores stage icons, connector lines, bars, labels, and spacing

The stacked browser smoke PR #344 adds coverage for the runtime/data/layout regressions missed by the existing CI checks.


paulbalaji:
The Zustand render loop is fixed in this PR: the preview now hydrates and renders the full application shell.

The Latest Messages feed is still showing its query-level error state, so the preview is not fully healthy yet. I am actively investigating that separate GraphQL/data-path regression and will update this PR with the root cause and any additional fix required before merge.

paulbalaji:
Update: the latest preview confirms both functional fixes are working — the app hydrates and Latest Messages loads again.

It is still **not merge-ready** because the Tailwind 4 build has a separate layout regression. Compared with the rolled-back production deployment, the same header logo classes (`h-8 w-auto sm:h-10`) compute to `1494 × 365px` instead of `164 × 40px`; the footer logo is affected similarly. This is why the header/footer expand over the page even though their source classes are unchanged.

I am tracing the Tailwind 4 generated CSS/cascade now and will update this PR with the minimal compatibility fix plus production-browser regression coverage.

paulbalaji:
Latest preview verification: https://hyperlane-explorer-ob6qm3jvo-abacus-works.vercel.app/

All three regressions are now resolved on the deployed head (`9c35a2d`):

- the app hydrates without the Zustand render loop or error boundary
- scraped domains and Latest Messages load through POST GraphQL requests
- the Tailwind 4 cascade fix restores the compact search bar, filters, message table, sticky header, and footer to the rolled-back production baseline

Vercel, typecheck, format, lint, security, and dependency checks are green. The GitHub Actions build is the only check still running; I am also preparing the stacked browser smoke-test PR described above.

troykessler:
the popup on origin and destination still looks broken as well as the message timeline on the message page
Screenshot 2026-08-10 at 12 24 36
Screenshot 2026-08-10 at 12 25 44

paulbalaji:
Homepage visual verification is complete on the deployed head `9c35a2d`.

I compared the [latest preview](https://hyperlane-explorer-ob6qm3jvo-abacus-works.vercel.app/) homepage side-by-side with the rolled-back [production Explorer](https://explorer.hyperlane.xyz/) at the same desktop viewport. The header/footer logos, search input and button, filter selectors, refresh button, table sizing, and page spacing now match the working homepage baseline, while hydration and live message loading remain healthy.

Follow-up review found two remaining Tailwind/widget regressions outside that homepage check: the Origin/Destination chain-picker popup and the message-detail timeline are still incorrectly styled. This PR is therefore not merge-ready yet. I am comparing those components against production and will update the fix and browser coverage before requesting merge.

All current CI checks, including the production build and Vercel deployment, are green; they do not presently exercise these deeper interactive/detail-page states.

paulbalaji:
Thanks @troykessler — confirmed both regressions and fixed them on the current head `192e38b`.

The partial Tailwind 4 workaround was the problem: lowering the entire precompiled widget stylesheet restored Explorer utilities but also lowered the widget's own `htw-*` utilities, which broke the chain picker and timeline. The PR now keeps the latest Tailwind 3 release (`3.4.19`) while retaining the rest of the dependency upgrade. A proper Tailwind 4 migration will follow after the browser smoke coverage lands.

Deployed verification: https://hyperlane-explorer-git-agent-fix-zustand-5-e98dc0-abacus-works.vercel.app

- Origin/Destination picker: search, controls, two-column chain rows, icons, and spacing match production
- supplied message timeline: stage icons, vertical connectors, chevrons, bars, labels, and spacing match production
- picker geometry matches production exactly (`512px` panel; `488px` grid with `234px / 234px / 0px` columns)
- timeline geometry/computed layout matches production exactly
- hydration, live Latest Messages, and POST GraphQL transport remain healthy
- no uncaught browser errors
- all CI and Vercel checks are green

The earlier visual-verification comment was edited to make clear that it only covered the homepage before these additional states were reported.

troykessler (approved):

PR #344, opened 2026-08-10
test: add Explorer production smoke coverage

paulbalaji:
## What

- add a Playwright production-browser smoke test for the Explorer homepage, chain picker, and message timeline
- run it against the built Next app in the existing CI build job
- verify hydration, live Latest Messages data, POST GraphQL transport, and app/widget layout

## Why

The dependency upgrade passed build, typecheck, lint, unit tests, and Vercel deployment while the deployed app had three user-visible regressions:

- a Zustand 5 render loop during hydration
- urql 5 GET requests rejected by the GraphQL edge
- Tailwind 4 utilities losing to the unlayered widget preflight

This test exercises the actual browser/runtime/data path that those checks missed.

## Assertions

- search UI hydrates without uncaught page errors or the app error boundary
- at least one live message row renders
- GraphQL queries use POST, not GET
- header/footer logos and search input keep their compact production heights
- Origin, Destination, Time, and Status controls remain compact
- the Origin chain picker renders its search, controls, and chain rows as the intended grids
- a delivered message renders four timeline bars and four correctly positioned stage icons
- Jest excludes the Playwright-only smoke directory

## Verification

- `pnpm build`
- `pnpm test:smoke` — 1 passed

Follows the merged runtime/style hotfix #343. This coverage is the gate for the separate Tailwind 4 migration follow-up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
   workflow_dispatch:
 
 jobs:
@@ -44,6 +43,12 @@ jobs:
 
       - name: build
         run: pnpm run build
+
+      - name: Install Chromium
+        run: pnpm exec playwright install --with-deps chromium
+
+      - name: Smoke test
+        run: pnpm run test:smoke
 
   typecheck:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


paulbalaji:
Exact-head verification complete at `d762650`:

- all CI checks pass, including the production build + Playwright smoke job
- local smoke covers hydration, live messages, POST transport, homepage controls, the chain picker, and the message timeline
- #345 now runs this same smoke suite as its stacked Tailwind 4 migration gate

PR #345, opened 2026-08-10
chore: upgrade Tailwind to 4

paulbalaji:
## What

- upgrade Tailwind CSS and its PostCSS plugin to 4.3.3
- declare the complete Tailwind layer order before importing either stylesheet
- place the precompiled Tailwind 3 widget bundle in the `components` layer
- restore Oxfmt Tailwind class sorting now that the project uses Tailwind 4 again

## Why

The first Tailwind 4 migration imported the widget bundle into `components` before Tailwind declared its native layers. CSS fixes layer precedence at the first declaration, so the effective order allowed later Tailwind base rules to override widget utilities. That broke the Origin/Destination picker and message timeline.

Declaring `theme, base, components, utilities` first gives the cascade a deterministic order:

- Tailwind and widget resets remain below component rules
- prefixed `htw-*` widget utilities retain their intended component styling
- Explorer utilities remain last and can override component defaults

## Browser gate

This PR is stacked on #344, which runs the built Next application in Chromium and covers every regression observed during the original migration:

- hydration completes without uncaught errors
- Latest Messages loads through POST GraphQL requests
- homepage logos, search, filters, and table remain compact
- the chain picker renders its controls and rows as grids
- the message timeline renders four bars and four correctly positioned stage icons

## Verification

- `pnpm build`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm run format` — no source changes
- `pnpm test:smoke` — 1 passed


paulbalaji:
Exact-head verification complete at `90a3376`:

- all CI and Vercel checks pass, including the stacked Playwright smoke suite
- the deployed Tailwind 4 preview loads live messages with no page errors
- homepage controls, chain picker layout, and message timeline pass the smoke assertions
- deployed picker/timeline geometry matches the known-good Tailwind 3 baseline (512px picker; 488px two-column rows; 56px timeline top spacing; correctly positioned 36×52px status icons)

Preview: https://hyperlane-explorer-git-agent-upgrade-tailwind-4-abacus-works.vercel.app

PR #346, opened 2026-08-11
fix: add env override for API URL

troykessler:
## Summary
- Adds NEXT_PUBLIC_API_URL to .env.example
- Reads NEXT_PUBLIC_API_URL before falling back to the existing hosted GraphQL endpoint

## Testing
- Not run; env/config-only change

### DIFF
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/dev/types/routes.d.ts";
+import "./.next/types/routes.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/pages/api-reference/config/typescript for more information.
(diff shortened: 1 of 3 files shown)


next-env.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/types/routes.d.ts";
+import "./.next/dev/types/routes.d.ts";
```
paulbalaji: Please revert this generated `next-env.d.ts` change. It is unrelated to the API URL override and switches the checked-in type reference to the dev-only `.next/dev` output; local Next commands can rewrite this file, and the file itself says not to edit it. Keeping the base version leaves this PR as the intended env/config-only change.

paulbalaji (commented):
reviewed exact head `8a2baa5e`. the `NEXT_PUBLIC_API_URL` override is scoped correctly and exact-head build/typecheck/lint/format are green. one unrelated generated-file drift inline; remove that and this is ready to approve.

Xaroz (approved):
need to run format it seems

PR #348, opened 2026-08-20
feat: add live scraper proxy updates

troykessler:
## Summary

- use one app-level scraper-proxy WebSocket across message list and detail views
- merge live message updates into GraphQL results and apply active filters locally
- use delivery-status polling only until the WebSocket is connected
- default GraphQL requests to explorer-api.hyperlane.xyz and rename the explorer route to /messages

## Validation

- pnpm run typecheck
- pnpm run lint
- pnpm run test --runInBand
- git diff main --check


src/features/messages/queries/useLiveMessages.ts
```diff
@@ -0,0 +1,206 @@
...
+    previousConnectionState.current = connectionState;
+  }, [connectionState, enabled]);
+}
+
+export function shouldRefreshAfterReconnect(
+  enabled: boolean,
+  previousConnectionState: ExplorerConnectionState,
+  connectionState: ExplorerConnectionState,
+) {
+  return enabled && previousConnectionState === 'disconnected' && connectionState === 'connected';
```
paulbalaji: Major: refresh on the initial `connecting -> connected` transition too. The initial GraphQL request and WebSocket subscription are not an atomic handoff: the live endpoint's `ready` frame carries no snapshot/cursor, so an upsert committed after the GraphQL snapshot but before readiness can belong to neither source. Once `connected`, both search/detail polling stop, leaving the row missing or stale for the session. Please refresh whenever an enabled consumer first becomes connected (or add a cursor/snapshot protocol), with a regression test for snapshot -> missed upsert -> ready.

src/features/messages/queries/useLiveMessages.ts
```diff
@@ -0,0 +1,206 @@
...
+        watchHeartbeat(socket);
+        const message = parseMessage(data);
+        if (message?.type === 'ready') {
+          reconnectDelayMs = INITIAL_RECONNECT_DELAY_MS;
+          setConnectionState('connected');
+        } else if (message?.type === 'message_upsert') {
+          setMessageRows((rows) =>
+            [message.data, ...rows.filter((row) => row.msg_id !== message.data.msg_id)].slice(
+              0,
+              MAX_RETAINED_MESSAGES,
```
paulbalaji: Major: this bounded global overlay is also the only live state for detail subscriptions. A pending GraphQL baseline can become Delivered from an upsert, then roll back to the stale Pending row after 500 unrelated upserts evict it. The socket remains connected, so both the GraphQL interval and on-chain delivery polling stay disabled until reconnect/reload. Please persist upserts into query/cache state, preserve active subscriptions separately, or retain a safety refresh; test pending baseline -> delivered upsert -> 500 unrelated upserts.

src/features/deliveryStatus/useMessageDeliveryStatus.tsx
```diff
@@ -98,7 +101,10 @@ export function useMessageDeliveryStatus({
     retry: false,
     refetchInterval: (query) =>
       query.state.data?.message.status === MessageStatus.Delivered ? false : 10_000,
-    enabled: enabled && !!multiProvider,
+    enabled:
+      enabled &&
+      !!multiProvider &&
+      shouldUseDeliveryStatusPolling(message.isPiMsg, explorerConnectionState),
```
paulbalaji: Major: this gates the entire query, not just its `refetchInterval`. For a connected non-PI pending/failing message, `fetchDeliveryStatus` never runs even once, so the detail page loses the on-chain failure reason, calldata, ISM/IGP details, and any faster delivery detection; WebSocket rows do not contain `debugResult`. Keep the initial check enabled and only suppress subsequent polling, or split live status from debugger enrichment.

paulbalaji (commented):
Reviewed exact head `e2df12a` with parallel Codex/Kimi adversarial passes, full source/history/thread reconciliation, and live endpoint probing. Local exact-head validation passed: 140/140 tests, typecheck, lint, production build, and diff check.

I found four major correctness gaps, posted inline. I also verified the existing unresolved WebSocket URL and external-payload validation threads; those remain valid and I have not duplicated them. This is a COMMENT review, not REQUEST_CHANGES.

src/features/messages/queries/useMessageQuery.ts
```diff
@@ -123,53 +199,92 @@ export function useMessageSearchQuery(
...
     isPendingFilter,
+    { cached: !liveSearchEnabled },
   );
 
   // Execute query
   const [result, reexecuteQuery] = useQuery<MessagesStubQueryResult>({
     query,
     variables,
-    pause: !isValidInput,
+    pause: !isValidInput || !isSearchMetadataReady,
```
paulbalaji, line 258: Major: `useScrapedDomains()` exposes `isError`, but this hook discards it. If `DOMAINS_QUERY` fails or returns no domains, this query stays paused, line 213 reports `isFetching=true`, and the returned `isError` only reflects the message query that never ran. The home/search UI therefore spins forever with no error or recovery. Please propagate the metadata-query error/readiness state and cover the failure path.

src/features/messages/MessageTable.tsx
```diff
@@ -272,12 +406,22 @@ export const MessageSummaryRow = memo(function MessageSummaryRow({
...
+          <span className="flex h-[18px] w-[18px] items-center justify-center" title={statusTitle}>
+            {statusIcon}
+          </span>
+        )}
       </LinkCell>
     </>
   );
 });
 
+const RelativeTime = memo(function RelativeTime({ timestamp }: { timestamp: number }) {
```
Xaroz, line 377: confused about this, I am not sure if it is worth it to create a full context for something that is called once

src/features/messages/MessageTable.tsx
```diff
@@ -69,37 +88,140 @@ export function MessageTable({
...
+                  chainMetadataResolver={chainMetadataResolver}
+                  router={router}
+                  warpRouteChainAddressMap={warpRouteChainAddressMap}
+                />
+              </tr>
+            );
+          })}
+        </tbody>
+      </table>
+      <style jsx global>{`
```
Xaroz: Can we avoid injecting global css from the component and also use tailwind setup if possible? Something like this would probably work inside tailwind.config

```
extend: {
  keyframes: {
    'live-message-insert': {
      '0%': {
        opacity: '0',
        transform: 'translateY(-10px)',
        boxShadow: 'inset 3px 0 0 #1a0a28',
      },
      '35%': {
        opacity: '1',
        transform: 'translateY(0)',
        boxShadow: 'inset 3px 0 0 #1a0a28',
      },
      '100%': {
        opacity: '1',
        transform: 'translateY(0)',
        boxShadow: 'inset 0 0 0 transparent',
      },
    },
  },
  animation: {
    'live-message-insert': 'live-message-insert 900ms ease-out',
  },
},
```

and then something like, which respect reduced motion preferences

`<tr className="motion-safe:animate-live-message-insert">`

src/features/messages/queries/useLiveMessages.ts
```diff
@@ -0,0 +1,206 @@
+import {
```
Xaroz: my nit is that there there seems to be multiple different hooks logic mixed here and would probably prefer this to be split into logical chunks

src/pages/_app.tsx
```diff
@@ -102,13 +103,15 @@ export default function App({ Component, router, pageProps }: AppProps) {
...
     <QueryClientProvider client={reactQueryClient}>
       <UrqlProvider value={urqlClient}>
-        <AppLayout pathName={router.pathname}>
-          {pendingRoute ? (
-            getRouteLoadingContent(pendingRoute) || <Component {...pageProps} />
-          ) : (
-            <Component {...pageProps} />
-          )}
-        </AppLayout>
+        <ExplorerEventsProvider>
```
Xaroz: Major: do not open this socket on routes with no live consumer. The provider connects immediately, so `/tx/[txHash]` and even 404s occupy Explorer capacity despite never using the live hooks. The scraper proxy hard-caps Explorer sockets at 400 ([source](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/9a8bb170b25732e0ce78b7f3f866b7179f47e077/typescript/scraper-proxy/src/live/event-websocket.ts#L28-L39)); once those slots are consumed, real search/detail clients are rejected and fall back to polling. Keep the context mounted if useful, but route-gate or reference-count the connection, with a regression test that a transaction route never constructs `WebSocket`.

src/features/messages/queries/useMessageQuery.ts
```diff
@@ -123,53 +199,92 @@ export function useMessageSearchQuery(
     dbStatusFilter,
     warpAddresses,
     isPendingFilter,
+    { cached: !liveSearchEnabled },
```
Xaroz: Major: cache mode needs to follow the actual WebSocket state, not metadata readiness. If the URL is unavailable, the server's 400 socket slots are full, or the socket disconnects, `isLiveConnected=false` resumes 15-second polling while this query remains uncached; the detail query also hard-codes `cached: false` at line 305 and polls every 10 seconds. The previous queries used `@cached(ttl: 5)`, and the proxy allows only 25 concurrent GraphQL requests ([source](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/9a8bb170b25732e0ce78b7f3f866b7179f47e077/typescript/scraper-proxy/README.md#L42-L48)). Preserve server caching for fallback polling and bypass/refresh it only for the connected handoff, with unavailable/disconnected regression cases.

src/features/messages/MessageTable.tsx
```diff
@@ -69,37 +88,140 @@ export function MessageTable({
...
+      insertedMessageTimers.current.forEach((timer) => clearTimeout(timer));
+      insertedMessageTimers.current.clear();
+      deliveredMessageTimers.current.forEach((timer) => clearTimeout(timer));
+      deliveredMessageTimers.current.clear();
+    },
+    [],
+  );
+
+  useEffect(() => {
+    const timer = setInterval(() => setTimeRefreshKey((key) => key + 1), TIME_SENT_REFRESH_MS);
```
Xaroz: Minor/refactor: use the existing `useVisibleInterval` here. It already wraps widgets' shared `useInterval` and suppresses callbacks while the tab is hidden; this raw one-second timer otherwise keeps scheduling timestamp state updates in background tabs. The context is still worthwhile because it limits each tick to the small `RelativeTime` consumers instead of re-rendering every memoized row.

Xaroz (commented):
## Review summary

Reviewed exact head `e2df12a` across all 15 changed files, relevant history/callers, the local SDK/utils/widgets sources, and the authoritative scraper-proxy implementation. I found two additional availability regressions, posted inline.

The existing four major correctness threads (initial handoff gap, 500-row detail eviction, disabled debugger/on-chain query, metadata-error infinite loading) and the WebSocket URL/payload validation threads remain valid; I did not duplicate them. I also agree with the current hook-separation and `motion-safe` animation threads.

## Regression and backend evidence

- The scraper proxy explicitly has no historical `/messages` catch-up, hard-caps Explorer sockets at 400, and caps GraphQL at 25 concurrent requests: [authoritative README](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/9a8bb170b25732e0ce78b7f3f866b7179f47e077/typescript/scraper-proxy/README.md#L39-L48), [server constants](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/9a8bb170b25732e0ce78b7f3f866b7179f47e077/typescript/scraper-proxy/src/live/event-websocket.ts#L28-L39).
- Live probing confirmed the deployed endpoint currently sends `ready`, 30-second heartbeats, and full `message_upsert` rows. `/graphql` accepted the PR's cached domain query.
- Historical review: search polling dates to `bd1e4f0`; delivery polling/debug enrichment dates to `b6baaa1`. Removing those safety nets therefore needs integration tests around connection transitions, not only pure helper tests.

## Shared-code/refactoring research

- `@hyperlane-xyz/utils`: the new protocol-aware filter correctly reuses `isAddress`, `eqAddress`, and transaction-hash validators. No SDK/utils WebSocket/reconnect abstraction exists, so Explorer ownership is reasonable.
- `@hyperlane-xyz/widgets`: `useInterval` and `SpinnerIcon` already exist. The visibility-aware local wrapper is called out inline; the existing pending-icon thread should use `SpinnerIcon` rather than retain the custom CSS spinner.
- SDK warp-route types/maps are already reused; I found no shared SDK matcher that replaces the domain-aware Explorer filter.

## Observation outside changed lines

- `src/features/messages/MessageTable.tsx:64-89`: live updates create a new `messageList` for every upsert, so the prefetch effect cancels/reschedules even when `backgroundPrefetchKey` is unchanged. Snapshot the top rows by that key (or keep them in a ref) so unrelated live events cannot starve navigation prefetch.
- The `RelativeTimeContext` itself is justified: one timer invalidates only the small consumers while the full rows stay memoized. Keep it, but make its timer visibility-aware as noted inline.

## Validation

- 140/140 tests passed
- typecheck passed
- lint passed
- production build passed (existing `styled-jsx/style.js` warning)
- `git diff --check` passed
- GitHub checks currently pass

Recommendation: address the existing correctness blockers plus the two availability findings before merge.

src/features/messages/queries/useLiveMessages.ts
troykessler: Addressed in ec8080f. Socket lifecycle, event validation, and provider/context now live in ExplorerEventsProvider.ts; consumer subscription/refresh hooks remain in useLiveMessages.ts.

src/consts/config.ts
```diff
@@ -1,6 +1,7 @@
 const isDevMode = process.env.NODE_ENV === 'development';
 const version = process.env.NEXT_PUBLIC_VERSION ?? null;
-const apiUrl = process.env.NEXT_PUBLIC_API_URL || 'https://explorer4.hasura.app/v1/graphql';
+const apiUrl = process.env.NEXT_PUBLIC_API_URL || 'https://explorer-api.hyperlane.xyz/graphql';
+const wsUrl = process.env.NEXT_PUBLIC_WS_URL ?? graphqlUrlToWsUrl(apiUrl);
```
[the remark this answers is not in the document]
troykessler: Fixed in ec8080f. NEXT_PUBLIC_WS_URL is now parsed and restricted to ws/wss URLs without fragments; invalid configuration logs and yields the unavailable state. Added malformed, unsupported-scheme, fragment, and valid URL tests.

src/features/messages/queries/useLiveMessages.ts
```diff
@@ -0,0 +1,208 @@
...
+function parseMessage(data: unknown): MessageUpsert | { type: 'heartbeat' | 'ready' } | null {
+  if (typeof data !== 'string') return null;
+  try {
+    const message = JSON.parse(data) as { type?: unknown };
+    if (message.type === 'ready' || message.type === 'heartbeat') return { type: message.type };
+    if (message.type === 'message_upsert') return message as MessageUpsert;
+    return null;
+  } catch {
+    return null;
+  }
```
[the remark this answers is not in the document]
troykessler: Fixed in ec8080f. Incoming message_upsert data is validated against the MessageEntry field contract and timestamp before retention; malformed events are logged and discarded. Added regression coverage.

src/features/messages/queries/useLiveMessages.ts
troykessler: Fixed in ec8080f. Detail subscriptions now register per-message listeners and retain their latest matching upsert independently of the bounded list overlay. Added retention and subscription-change regression coverage.

src/features/messages/MessageTable.tsx
troykessler: Fixed in ec8080f. The keyframes/animation now live in Tailwind config and the row uses motion-safe:animate-live-message-insert.

src/features/messages/queries/useLiveMessages.ts
troykessler: Fixed in ec8080f. Enabled consumers now refresh on every non-connected to connected transition, including the initial handoff. Updated the regression test accordingly.

src/features/messages/MessageTable.tsx
```diff
@@ -69,37 +88,140 @@ export function MessageTable({
...
+            const isInserted = insertedMessageIds.has(m.id);
+            const isDelivered = deliveredMessageIds.has(m.id);
+            return (
+              <tr
+                key={`message-${m.id}`}
+                className={`border-primary-50 hover:bg-accent-50 active:bg-accent-100 relative cursor-pointer border-b last:border-0 ${
+                  isFetching && 'blur-xs'
+                } ${
+                  isInserted ? 'bg-primary-50 live-message-insert' : ''
+                } ${isDelivered ? 'bg-[#dcfce7]' : ''} transition-all duration-500`}
```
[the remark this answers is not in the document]
troykessler: Fixed in ec8080f. The delivered highlight now uses the shared bg-green-100 token.

src/features/messages/queries/useLiveMessages.ts
```diff
@@ -0,0 +1,208 @@
...
+  const refreshRef = useRef(refresh);
+  const previousConnectionState = useRef(connectionState);
+  refreshRef.current = refresh;
+
+  useEffect(() => {
+    if (shouldRefreshAfterReconnect(enabled, previousConnectionState.current, connectionState)) {
+      refreshRef.current();
+    }
+    previousConnectionState.current = connectionState;
+  }, [connectionState, enabled]);
```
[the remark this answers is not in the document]
troykessler: Fixed in ec8080f. The refresh ref update now runs in an effect before the connection-transition effect.

src/features/messages/queries/useMessageQuery.ts
troykessler: Fixed in ec8080f. Search and detail GraphQL queries now retain @cached during unavailable, connecting, and disconnected states, bypassing it only while the WebSocket is actually connected. Added coverage for all connection states.

src/features/messages/queries/useMessageQuery.ts
troykessler: Fixed in ec8080f. Search now propagates scraped-domain/chain query errors and treats a completed empty domains response as an error instead of an endless loading state. Added both failure-path tests.

src/features/deliveryStatus/useMessageDeliveryStatus.tsx
troykessler: Fixed in ec8080f. The delivery-status query remains enabled for its initial debugger/on-chain enrichment; only its subsequent interval is disabled while a non-PI live connection is active. Added interval behavior coverage.

src/pages/_app.tsx
troykessler: Fixed in ec8080f. The provider stays mounted but receives a route gate; only search and message-detail routes connect. Added a regression test rendering the transaction-route configuration and asserting WebSocket is never constructed.

src/features/messages/MessageTable.tsx
troykessler: Kept the context intentionally: the value is consumed by every RelativeTime cell, so the one shared tick updates only those small consumers while memoized rows remain stable. I did switch the ticker to useVisibleInterval so hidden tabs do no work.

src/features/messages/MessageTable.tsx
troykessler: Fixed in ec8080f. The shared relative-time tick now uses useVisibleInterval, so timestamp updates pause while the tab is hidden.

src/features/messages/queries/useMessageQuery.ts
troykessler: Follow-up correction in 89e24da: the first readiness fix could classify the initial empty store as a completed empty response, briefly showing the generic error. Metadata now distinguishes not-run from completed-empty and exposes fresh query data immediately while the store-sync effect runs. Added initial-not-run coverage.

src/features/messages/queries/useLiveMessages.ts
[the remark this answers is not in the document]
troykessler: Follow-up correction in 89e24da: the first validation fix was too strict because the live proxy serializes database numeric/bigint fields as strings and pending rows omit destination fields. I reproduced this against wss://explorer-api.hyperlane.xyz/messages, updated validation to the actual wire format, and added pending + delivered proxy-payload regression tests.

src/features/messages/queries/useMessageQuery.ts
```diff
@@ -26,6 +38,23 @@ export function isValidSearchQuery(input: string) {
...
+  scrapedDomainCount: number,
+  mainnetDomainCount: number,
+  isFetching: boolean,
+  isError: boolean,
+) {
+  const isReady = scrapedDomainCount > 0 && mainnetDomainCount > 0;
+  return {
+    isReady,
+    isError: isError || (!isFetching && scrapedDomainCount === 0),
+  };
```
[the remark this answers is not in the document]
troykessler: Fixed in 9baa672. Completed metadata with zero mainnet domains now reports an error. Completion is gated on both the domains query and chain metadata load, avoiding the initial-load error regression. Added coverage.

src/consts/config.ts
```diff
@@ -30,3 +35,27 @@ export const config: Config = Object.freeze({
...
 export const debugIgnoredChains = ['treasure', 'treasuretopaz'];
+
+export function resolveWsUrl(override: string | undefined, graphqlUrl: string): string | null {
+  const url = override ?? graphqlUrl;
+  try {
+    const parsed = new URL(url);
+    if (override) {
+      if (!['ws:', 'wss:'].includes(parsed.protocol) || parsed.hash) {
+        throw new Error('NEXT_PUBLIC_WS_URL must use ws/wss and must not contain a fragment');
+      }
```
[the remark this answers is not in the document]
troykessler: Fixed in 9baa672. Blank overrides now normalize to unset before URL selection/validation. Added empty-override fallback and http-to-ws derivation coverage.

paulbalaji (commented):
Full rereview of exact head `f0b4b9f`. I verified all prior threads and fixes, reviewed the full base diff plus the five-commit follow-up delta, and ran 155/155 tests, typecheck, lint, production build, diff check, and a live WebSocket payload probe. Current CI is green and all existing threads are resolved.

The previous four correctness findings are fixed. I found two remaining lifecycle gaps in the retention strategy, posted inline. This is a non-blocking COMMENT review, not REQUEST_CHANGES.

src/features/messages/queries/useLiveMessages.ts
```diff
@@ -0,0 +1,102 @@
...
+  return { connected: enabled && connectionState === 'connected', messageRow };
+}
+
+export function selectSubscribedMessageRow(
+  currentMessageRow: MessageEntry | null,
+  retainedMessageRow: MessageEntry | null,
+  messageId: string,
+) {
+  const retainedRowMatches = normalizeId(retainedMessageRow?.msg_id) === normalizeId(messageId);
+  return currentMessageRow ?? (retainedRowMatches ? retainedMessageRow : null);
```
paulbalaji, line 74: Major: clear the retained row when the socket leaves `connected`. This value survives `scheduleReconnect()` clearing `messageRows`. If a retained Pending upsert is followed by delivery while disconnected, polling/refetch can load Delivered into `messageList`, but `liveMessage ?? messageList[0]` still prefers this Pending row; the reconnect refresh is shadowed too. Keying by ID only covers subscription changes, not connection epochs. Please test Pending upsert → disconnected → Delivered GraphQL result.

src/features/messages/queries/ExplorerEventsProvider.ts
```diff
@@ -0,0 +1,207 @@
...
+          setConnectionState('connected');
+        } else if (message?.type === 'message_upsert') {
+          const normalizedId = normalizeId(message.data.msg_id);
+          if (normalizedId) {
+            listenersRef.current.get(normalizedId)?.forEach((listener) => listener(message.data));
+          }
+          setMessageRows((rows) =>
+            [message.data, ...rows.filter((row) => row.msg_id !== message.data.msg_id)].slice(
+              0,
+              MAX_RETAINED_MESSAGES,
```
paulbalaji, line 128: Major: this cap still drops live-only rows from filtered searches. The new per-ID listener preserves detail subscriptions, but `useLatestMessageRows` still consumes only this 500-row overlay. A matching result received after the last GraphQL snapshot disappears after 500 unrelated upserts, and connected mode suppresses the 15s query poll, so it stays missing until manual refresh/reconnect. Persist upserts into query/cache state or keep a bounded connected safety refresh; test a filtered live-only row plus 500 unrelated events.

src/features/messages/queries/useLiveMessages.ts
troykessler: Fixed in 6b1e91d. Retained detail rows are cleared and ignored whenever the socket leaves connected, so fallback GraphQL data is authoritative during disconnects and reconnect refreshes cannot be shadowed. Added Pending-retained → disconnected regression coverage.

src/features/messages/queries/ExplorerEventsProvider.ts
troykessler: Fixed in 6b1e91d. Filtered searches keep the existing 15-second safety refresh even while live updates are connected; only the unfiltered latest feed becomes WebSocket-only. Thus eviction can no longer leave filtered live-only results missing indefinitely. Added polling-policy coverage.

paulbalaji (commented):
Full rereview of exact head `6b1e91d`: full base diff and follow-up delta, author replies, CI, and all 21 review threads reconciled. Exact-head validation passed: 157/157 tests, typecheck, lint, production build, and diff check.

The two findings from my prior review are correctly fixed, with regression coverage. I found no new issue in the follow-up delta. I independently verified the remaining unresolved CodeRabbit finding at `useScrapedChains.ts:57`: a rejected registry metadata load is not represented in state, so the new readiness gate keeps the message query paused with `isFetching=true` indefinitely. The existing inline thread already covers it, so I have not duplicated it.

This is a non-blocking COMMENT review, not REQUEST_CHANGES.

src/features/messages/queries/useMessageQuery.ts
```diff
@@ -110,7 +209,20 @@ export function useMessageSearchQuery(
...
+    hasInput ||
+    originChainNameFilter ||
+    destinationChainNameFilter ||
+    startTimeFilter ||
+    endTimeFilter ||
+    statusFilter !== 'all' ||
+    warpRouteAddresses.length > 0
+  );
+  const liveDomains = useMemo(
+    () => (hasFilters ? [] : mainnetDomainIds),
```
Xaroz: **[348-001] Preserve mainnet scope for status/time-only filters**

With `?status=pending` on a clean preview load, 15 of the first 20 visible rows were `Base Sepolia → Unknown`. Production already exhibits this, so it is not a new regression, but returning `[]` here makes `useLatestMessageRows` accept every socket domain and preserves the issue in the new live path.

Could we keep `mainnetDomainIds` for generic status/time filters, and only widen scope when the user explicitly selects a testnet chain (or when an identifier search intentionally needs cross-environment results)? A regression test with a mainnet and Base Sepolia upsert under `status=pending` would make the intended policy explicit.

src/features/messages/MessageTable.tsx
```diff
@@ -69,37 +89,114 @@ export function MessageTable({
...
+    () => () => {
+      insertedMessageTimers.current.forEach((timer) => clearTimeout(timer));
+      insertedMessageTimers.current.clear();
+      deliveredMessageTimers.current.forEach((timer) => clearTimeout(timer));
+      deliveredMessageTimers.current.clear();
+    },
+    [],
+  );
+
+  useVisibleInterval(() => setTimeRefreshKey((key) => key + 1), TIME_SENT_REFRESH_MS);
```
Xaroz: **[348-002] Avoid refreshing every row every second**

This updates every `RelativeTime` consumer once per second while the tab is visible. In the preview I observed it on a 229-row Pending result, so the table continuously reconciles hundreds of time cells even with no socket event. `useVisibleInterval` solves hidden-tab work, but not visible-tab cost.

Could the cadence be adaptive (1s only for very recent rows, then 10–60s), or otherwise limited to visible/recent rows? Please profile the large Pending view on a low-end mobile target.

src/features/messages/MessageTable.tsx
```diff
@@ -144,6 +241,18 @@ export const MessageSummaryRow = memo(function MessageSummaryRow({
         className="pt-px"
       />
     );
+  } else if (status === MessageStatus.Pending) {
+    statusTitle = 'Pending';
+    statusIcon = (
+      <span
+        className="block h-3.5 w-3.5 animate-spin rounded-full border-[1.5px]"
```
Xaroz: **[348-003] Respect reduced-motion for the Pending spinner**

The row insertion animation is `motion-safe`, but this new spinner always animates. Please use `motion-safe:animate-spin` and leave the ring static for reduced-motion users (or provide an equivalent non-animated Pending state).

Xaroz (commented):
Browser-tested the Vercel preview at desktop and mobile.

The socket flow itself worked: the feed received a new row without refresh, a Pending detail updated to Delivered, search/filter/detail/deep-link flows passed, and I saw no browser console errors.

I found three non-blocking follow-ups below. The Base Sepolia behavior also reproduces in production, so I am not treating it as a regression introduced by this PR; the new live-message merge preserves it.

src/features/messages/queries/useMessageQuery.ts
troykessler: Fixed in 064d726. Generic status/time filters now retain the mainnet domain scope. Identifier searches and explicitly selected testnet domains widen the live scope. Added regression coverage proving a Base Sepolia upsert is excluded from a mainnet-scoped pending search.

src/features/messages/MessageTable.tsx
troykessler: Fixed in 064d726. Relative timestamps now refresh independently with an age-adaptive cadence: 1s only for the first minute, 60s up to one hour, hourly up to one day, then no timer once the display is date-only. Hidden-tab suppression remains. Added cadence tests.

src/features/messages/MessageTable.tsx
troykessler: Fixed in 064d726. The Pending spinner now uses motion-safe:animate-spin, leaving a static ring when reduced motion is enabled.

src/features/chains/queries/useScrapedChains.ts
```diff
@@ -45,7 +48,13 @@ export function useScrapedChains(chainMetadataResolver: ChainMetadataResolver) {
     return { chains: scrapedChains };
   }, [chainMetadataResolver, chainMetadata, scrapedDomains]);
 
-  return { chains, isFetching, isError };
+  return {
+    chains,
+    scrapedDomains,
+    isFetching,
+    isError,
+    hasRun: hasRun && isChainMetadataLoaded,
```
paulbalaji, line 67: Major: this remains unfixed at the current head even though the earlier thread is resolved. `ensureChainMetadata()` still rejects without storing an error, so `isChainMetadataLoaded` remains false; this `hasRun` stays false, `getSearchMetadataState` reports neither ready nor error, and the message query remains paused with `isFetching=true` forever. Please track/expose the registry metadata error and test a rejected load.

paulbalaji (commented):
Full rereview at exact head `064d726fcc37790010b8a165a12b3eacac9cb9a4`.

The latest fixes correctly preserve mainnet scope for generic status/time filters, make relative-time refresh cadence age-dependent, and honor reduced-motion for the Pending spinner.

Validated the full 23-file PR and latest 8-file delta: 162/162 Jest tests, typecheck, lint, production build, `git diff --check`, preview HTTP, and live WebSocket readiness. Current CI/Vercel checks are green.

One existing Major remains: the chain-metadata rejection path can still leave message search loading forever. Its prior thread is marked resolved, but the underlying code path is unchanged, so I have re-raised it inline on the current diff.

src/features/messages/queries/useMessageQuery.ts
```diff
@@ -110,7 +218,24 @@ export function useMessageSearchQuery(
...
+    statusFilter !== 'all' ||
+    warpRouteAddresses.length > 0
+  );
+  const liveDomains = useMemo(() => {
+    const selectedDomainIds = [
+      originDomainId,
+      destDomainId,
+      ...warpRouteDomainAddresses.map((a) => a.domainId),
+    ].filter((domainId): domainId is number => domainId !== null);
+    return getLiveMessageDomainIds(mainnetDomainIds, hasInput, selectedDomainIds);
```
Xaroz: Major: apply this environment-scope decision to the GraphQL source too. This line constrains only `liveMessageRows`; `buildMessageSearchQuery` still sees `status=pending` (or a time-only filter) as `hasFilters=true`, so `buildDomainIdWhereClause` returns an empty origin/destination predicate and every 15-second safety refresh merges testnets back in. A live probe of the same 500-row Pending baseline returned 192 pending rows, 179 from Base Sepolia (`84532 -> 1337090`). Derive one shared scope policy for both the generated query and `useLatestMessageRows`, keeping identifier searches/explicit testnet selections unbounded, and add a generated-query regression test for the status-only case.

Xaroz (commented):
## Rereview summary

Reviewed exact head `064d726fcc37790010b8a165a12b3eacac9cb9a4`: full 23-file diff, all 25 review threads, the seven follow-up commits since my prior review, the Aug 25 delta, relevant history, and local Hyperlane SDK/utils/widgets sources.

One additional Major is inline: the latest fix scopes generic filters to mainnet only for WebSocket rows. The GraphQL half of the merged result still removes both mainnet predicates for any status/time filter. A live proxy probe of the 500-row baseline used by Pending returned 192 pending rows, 179 from Base Sepolia, so the composed UI behavior remains.

## Prior-thread reconciliation

- The Aug 25 mainnet-live-scope, adaptive-time, and reduced-motion changes are locally correct in isolation.
- The existing registry-metadata rejection Major remains valid and is already open at `useScrapedChains.ts:56`; I did not duplicate it.
- The other connection handoff, retention, polling/debugger, routing, validation, caching, and animation threads now have matching code and regression coverage. No additional variants survived caller/state-transition review.

## Shared-code/refactoring check

- Local widgets' `useInterval` already supports `number | null`; the updated `useVisibleInterval` correctly reuses it.
- Local widgets' `SpinnerIcon` hard-codes animation and has no reduced-motion-safe variant, so keeping the static reduced-motion ring locally is defensible.
- Local SDK/utils/widgets have no reusable WebSocket lifecycle/reconnect abstraction or relative-time cadence helper. Existing protocol-aware SDK/utils helpers are already reused.

## Validation

- 162/162 Jest tests passed
- typecheck passed
- lint passed
- production build passed
- `git diff origin/main...HEAD --check` passed
- current CI/Vercel checks are green
- live GraphQL probe reproduced the remaining scope gap

Recommendation: fix the open registry-error path and make GraphQL/live environment scope share one policy before merge.

src/features/messages/queries/useMessageQuery.ts
```diff
@@ -119,57 +244,95 @@ export function useMessageSearchQuery(
...
-    pause: !isValidInput,
+    pause: !isValidInput || !isSearchMetadataReady,
+    requestPolicy: 'cache-and-network',
   });
-  const { data, fetching: isFetching, error } = result;
+  const { data, fetching, error } = result;
+  const isFetching = isValidInput && !isSearchMetadataError && (!isSearchMetadataReady || fetching);
+  const refresh = useCallback(() => {
+    if (!query || !isValidInput) return;
+    reexecuteQuery({ requestPolicy: 'network-only' });
```
Xaroz, line 271: Major: make Refresh retry failed search prerequisites. The new metadata-error path leaves `isSearchMetadataReady=false`, so the operation above is paused. The page's Refresh button calls this callback, but it only reexecutes that paused message query; it never retries `ensureChainMetadata()`, and `useScrapedDomains()` likewise exposes no reexecutor. A transient registry or domains failure therefore remains unrecoverable until remount/reload even though Refresh is available. Please return a search-level retry that reruns failed prerequisites before the message query, with reject → Refresh → success coverage.

Xaroz (commented):
## Rereview summary

Reviewed exact head `f19bad76a770bdaf52945ad47d4848eb8ded8044`: full 25-file diff, the delta from the previously reviewed `064d726`, all 26 review threads, caller/state-transition paths, relevant history, adjacent local Hyperlane SDK/widgets/utils, and the current live proxy payload.

The two prior blockers are fixed:

- Registry metadata rejection now reaches the search error state and has regression coverage.
- `searchDomainIds` now drives both GraphQL and live-row scope; status/time-only searches stay mainnet-scoped while identifier or explicit-testnet searches widen.

One new Major is inline: the visible Refresh control cannot retry either failed search prerequisite, so a transient registry or scraped-domain failure still requires remount/reload.

## Review-state reconciliation

- All 26 existing review threads are resolved at this head.
- No new reviewer feedback was posted after the previous review.
- The WebSocket wire probe confirmed the live proxy still sends the full pending row used by both stub and detail parsing.

## Shared-code/refactoring check

- No matching environment-scope or metadata-error abstraction exists in the adjacent local SDK, widgets, or utils packages; these policies remain Explorer-specific.
- The adaptive timer correctly reuses widgets' existing `useInterval(number | null)` surface.

## Validation

- 170/170 Jest tests passed
- typecheck passed
- lint passed
- production build passed
- `git diff origin/main...HEAD --check` passed
- current CI/Vercel checks are green

Recommendation: wire Refresh to retry failed prerequisites and add recovery coverage before merge.

paulbalaji (approved):
Rereviewed exact head `9c31f599`. The prior Refresh finding is fixed: failed search prerequisites are retried, and successful metadata state unpauses the message query. I found no remaining material issues.

Verified locally: 171/171 tests, typecheck, lint, and `git diff --check`. Current GitHub CI and Vercel checks are green.

PR #351, opened 2026-08-26
fix(messages): keep loading state during filter searches

paulbalaji:
## Summary

- wait for results from the current filter request before showing rows or the empty state
- keep genuinely empty results stable during same-query background refreshes
- mount the first filtered result batch as the table baseline so historical rows do not get the live-message insertion animation
- cover current-request matching and background-refresh behavior with regression tests

## Test plan

- `pnpm test --runInBand`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm build`

### DIFF
--- a/src/features/messages/MessageSearch.tsx
+++ b/src/features/messages/MessageSearch.tsx
@@ -220,7 +220,7 @@ export function MessageSearch() {
   const isAnyFetching = isFetching || piSearchState.isFetching;
   const isAnyError = isError || piSearchState.isError;
   const hasAllRun = hasRun && (!shouldRunPiSearch || piSearchState.hasRun);
-  const isReadyToShowEmptyState = hasAllRun && isChainMetadataReady;
+  const isSearchSettled = hasAllRun && isChainMetadataReady;
   const isAnyMessageFound = isMessagesFound || piSearchState.isMessagesFound;
   const messageListResult = isMessagesFound ? messageList : piSearchState.messageList;
   const messageTableKey = [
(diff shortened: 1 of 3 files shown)


src/features/messages/MessageSearch.tsx
```diff
@@ -381,29 +381,22 @@ export function MessageSearch() {
...
           isPiFetching={piSearchState.isFetching}
         />
         <SearchEmptyError
           show={
-            !redirectUrl &&
-            !isAnyError &&
-            isValidInput &&
-            !isAnyMessageFound &&
-            isReadyToShowEmptyState
+            !redirectUrl && !isAnyError && isValidInput && !isAnyMessageFound && isSearchSettled
```
Xaroz, line 399: might need to run format?

haggis-hyperlane:
Gating the empty state on `!isAnyFetching` means a genuinely-empty filtered result flips to the loading spinner during each 15s background poll. `shouldPollMessageSearch` is always true when filters are set, so `isFetching` toggles true every interval, briefly hiding the "no results" state and showing the spinner.

If that flicker matters, gate the empty state on "data matches the current filters" instead of raw fetching (e.g. only treat the query as unsettled while the in-flight variables differ from the ones that produced the current data). Otherwise this is good to merge.

Xaroz (approved):

src/features/messages/MessageSearch.tsx
```diff
@@ -381,29 +381,22 @@ export function MessageSearch() {
...
+        )}
         <SearchFetching
-          show={
-            !isAnyError &&
-            isValidInput &&
-            !isAnyMessageFound &&
-            (!hasAllRun || !isChainMetadataReady)
-          }
+          show={!isAnyError && isValidInput && !isAnyMessageFound && !isSearchSettled}
           isPiFetching={piSearchState.isFetching}
```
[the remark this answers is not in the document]
paulbalaji: Fixed in f76569f. The loading state now depends on valid input, no error, and the current request being unsettled; preserved matching rows no longer leave a blank card.

src/features/messages/MessageSearch.tsx
paulbalaji: Checked the current head with `pnpm exec oxfmt --check` on all changed files; it passes with no formatting changes needed.

PR #355, opened 2026-08-27
feat: explorer history

troykessler:
## Summary
- add stable before/after cursor pagination for message searches
- keep displayed messages newest-first while paging by database ID
- restore the live first page when backward pagination reaches the newest batch
- reset pagination when search filters change

## Deployment
Deploy scraper-proxy query cursor support before this Explorer change.

## Validation
- pnpm run typecheck
- pnpm run lint
- pnpm run test --runInBand (181 tests)


src/features/messages/queries/useMessageQuery.ts
```diff
@@ -296,15 +343,19 @@ export function useMessageSearchQuery(
...
+    () => mergeMessageLists(liveMessageList, unfilteredMessageList, isBeforeQuery),
+    [isBeforeQuery, liveMessageList, unfilteredMessageList],
   );
   // Apply client-side filters. Note: these run after the DB LIMIT, so they
   // can shrink a page below the requested size — acceptable here since the
   // alternative (per-chain DB clauses) bloats the query and the pending
   // filter already relies on client-side narrowing.
-  const messageList = useMemo(() => {
-    let list = mergedUnfilteredMessageList;
+  const candidateMessageList = useMemo(
```
paulbalaji, line 390: **Medium — valid GraphQL matches can be evicted before filtering.** `mergedUnfilteredMessageList` includes up to 500 retained live rows that are scoped only by domains, but this truncates the merge to 51 before the address/hash/chain/status/warp filters below run. With 51 newer unrelated live rows, an older exact result returned by GraphQL is dropped here, so the search can show empty/PI fallback and miss its redirect. Filter live rows before merging, or filter the full merge while deriving the database continuation boundary separately. Please add a regression with 51 unrelated live rows plus one older matching GraphQL row.

src/features/messages/queries/useMessageQuery.ts
```diff
@@ -26,13 +32,34 @@ import {
...
+export function getMessagePage(
+  filteredMessages: MessageStub[],
+  candidateMessages: MessageStub[],
+  queryLimit: number,
+) {
+  const pageMessages = filteredMessages.slice(0, MESSAGE_PAGE_SIZE);
+  const continuationCursor =
+    filteredMessages.length > MESSAGE_PAGE_SIZE
+      ? pageMessages[pageMessages.length - 1]?.id || null
+      : candidateMessages.length === queryLimit
```
paulbalaji: **Medium — one parse failure can terminate history early.** This continuation check uses the parsed candidate count, while `queryResult()` logs and drops rows that fail parsing. A full 51-row database batch with one malformed/unsupported row becomes 50 candidates here, making `nextCursor` null although older rows exist. Preserve the raw returned count and boundary ID through parsing (or return explicit page metadata), and cover a full raw batch containing one rejected row.

src/features/messages/queries/parse.ts
```diff
@@ -61,17 +62,26 @@ function queryResult<D, M extends MessageStub>(
...
+  ascending = false,
 ) {
   if (!data || !Object.keys(data).length) return [];
   return deduplicateMessageList(
     Object.values(data)
       .flat()
       .map((d) => parseFn(chainMetadataResolver, scrapedChains, d))
       .filter((m): m is M => !!m)
-      .sort((a, b) => b.origin.timestamp - a.origin.timestamp),
+      .sort((a, b) =>
```
paulbalaji: **Low — this changes the shared parser's default ordering outside pagination.** These helpers previously returned send-time-descending results; they are also used by the public message/search/status API handlers and ICA/transaction paths. The paginated list re-sorts its visible page, but those other callers now silently inherit database-ID ordering. Please preserve timestamp ordering as the default and opt the cursor path into ID ordering explicitly, or document and test the broader API contract change.

paulbalaji (commented):
Reviewed exact head `2a2926aee9f36840df51d7aaea53150a549d9e16`. Posting COMMENT rather than approval because the live-row truncation can make valid searches return empty; two lower-severity pagination/API-contract findings are inline too.

I independently confirmed the existing unresolved CodeRabbit PI-state finding and did not duplicate it. I also treated the PR's documented deployment ordering as a hard release gate: production currently rejects `message_view(cursor: ...)`, and scraper-proxy PR hyperlane-monorepo#9341 remains open at `d0a45b1`, so that support must be merged and deployed first.

Validation: focused Jest 34/34, full Jest 23 suites / 192 tests, typecheck, Oxlint, and `git diff --check` passed locally; all completed required exact-head GitHub checks passed. Interactive preview inspection was unavailable in this session.

src/features/messages/queries/useMessageQuery.ts
troykessler: Fixed in 51f9381. Live and GraphQL rows are client-filtered before merge/truncation, so unrelated retained live rows cannot evict a valid GraphQL match. Added the requested 51-live-row regression.

src/features/messages/MessageSearch.tsx
```diff
@@ -368,26 +428,40 @@ export function MessageSearch() {
...
+              )}
+              <MessagePagination
+                isFirstPage={isFirstPage}
+                canGoPrevious={!!previousCursor}
+                canGoNext={!!nextCursor}
+                loading={isAnyFetching}
+                onFirstPage={resetPagination}
+                onPreviousPage={goToPreviousPage}
+                onNextPage={goToNextPage}
+              />
```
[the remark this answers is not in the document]
troykessler: Fixed in 51f9381. PI-derived fetching, errors, result presence, displayed rows, and loading state are now gated to the first page; cursor pages use only GraphQL pagination state.

src/features/messages/queries/useMessageQuery.ts
troykessler: Fixed in 51f9381. Cursor continuation and reverse boundaries now come from deduplicated raw GraphQL row IDs, independent of parsing success. Added a full 51-row batch regression with only 50 parsed rows.

src/features/messages/queries/parse.ts
troykessler: Fixed in 51f9381. Shared parsers again default to send-time-descending ordering. Only the paginated message-search path explicitly opts into ID ASC/DESC ordering.

src/features/messages/MessageSearch.tsx
```diff
@@ -368,31 +429,45 @@ export function MessageSearch() {
...
           </div>
         </div>
         {isSearchSettled && (
-          <Fade show={showMessageTable && !redirectUrl}>
-            <MessageTable
-              key={messageTableKey}
-              messageList={messageListResult}
-              isFetching={isAnyFetching}
-            />
+          <Fade show={(showMessageTable || !isFirstPage) && !redirectUrl}>
```
paulbalaji: **Medium — an empty filtered first page hides reachable older results.** A full raw batch can produce `messages=[]` with a non-null `nextCursor` (for example, the newest 500 rows are delivered but older pending rows exist). On page one, `showMessageTable` is false, so this hides the entire pagination subtree and `SearchEmptyError` leaves no way to follow that cursor. Please expose a continuation state (without the terminal empty error), or auto-advance empty raw pages, and cover the zero-match/full-raw-page case.

src/features/messages/queries/useMessageQuery.ts
```diff
@@ -114,6 +182,33 @@ export function messageMatchesSearchFilters(
...
+
+export function getMessageCandidates(
+  liveMessages: MessageStub[],
+  queryMessages: MessageStub[],
+  queryLimit: number,
+  filters: MessageSearchFilters,
+  warpRouteDomainAddresses: Array<{ domainId: number; address: string }>,
+  ascending = false,
+): MessageStub[] {
+  const matchingLiveMessages = filterMessageList(liveMessages, filters, warpRouteDomainAddresses);
```
paulbalaji: **Medium — merge before filtering so live status overrides stale query state.** With `status=pending`, a delivered live upsert is filtered out here while its same-ID stale pending GraphQL row survives below, so the live update cannot replace it and the message remains shown pending until polling catches up. Merge the full live/query lists first, then apply client filters and the limit; this still prevents unrelated live rows from evicting matches. Please add the delivered-live/pending-query transition case.

paulbalaji (commented):
Rereviewed exact head `51f93810d85d63eceba9500e4d66dbf3393f8dab`. The four prior threads are resolved and their fixes/tests check out. I found two remaining client-side correctness cases, posted inline; COMMENT only, not a change request.

Validation: focused Jest 36/36, full Jest 23 suites / 194 tests, TypeScript, Oxlint, and `git diff --check` passed locally. Current GitHub build/typecheck/format/lint/security/Socket/CodeRabbit/Vercel checks pass. Production Explorer API now executes exclusive ASC/DESC cursor queries correctly.

src/features/messages/queries/useMessageQuery.ts
troykessler: Fixed in 8bb1f61. Live rows now override query rows by message ID before client filtering and truncation, preserving fresh delivery status without allowing unrelated rows to evict matches. Added the delivered-live/stale-pending regression.

src/features/messages/MessageSearch.tsx
troykessler: Fixed in 8bb1f61. A settled page with a continuation cursor now renders pagination even with zero filtered matches, and the terminal empty state is suppressed while older results remain reachable. Added the zero-match/full-raw-page regression.

src/features/messages/MessageSearch.tsx
```diff
@@ -58,6 +59,19 @@ enum MESSAGE_QUERY_PARAMS {
   START_TIME = 'startTime',
   END_TIME = 'endTime',
   STATUS = 'status',
+  AFTER = 'after',
+  BEFORE = 'before',
+}
+
+export function parseMessageCursor(value: string): string | null {
```
Xaroz, line 68: **Low — reject cursors outside the backend `bigint` range.** This accepts arbitrarily long digit strings, but the deployed proxy binds the value against PostgreSQL `bigint`. A URL such as `?after=999999999999999999999999999999999999` currently returns `value ... is out of range for type bigint`; with `data: null`, the page shows the unknown-error state and no pagination control to clear the cursor. Please bound the parsed value to a valid positive signed-64-bit ID, clear invalid params, and add an oversized-cursor regression test.

src/features/messages/MessageSearch.tsx
```diff
@@ -434,6 +514,56 @@ export function MessageSearch() {
...
+  onFirstPage: () => void;
+  onPreviousPage: () => void;
+  onNextPage: () => void;
+}) {
+  return (
+    <nav
+      className="flex items-center justify-center gap-2 px-4 py-4 sm:gap-4"
+      aria-label="Message pages"
+    >
+      <button
```
Xaroz: **Non-blocking refactor — reuse the shared button primitive.** The adjacent widgets package already exports `Button` from [`typescript/widgets/src/components/Button.tsx`](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/main/typescript/widgets/src/components/Button.tsx), and this file already imports widgets components. Using it for these three controls would centralize button type/disabled/active behavior; a small shared class/config map can remove the repeated styling. I found no existing pagination component in the adjacent SDK, widgets, or utils trees.

Xaroz (commented):
Reviewed exact head `098adaff8a30757b323da1d107d2b7416d7d9dbd`, including all three follow-up commits. Posting COMMENT with one low input-boundary issue and one non-blocking reuse suggestion inline.

## Follow-up review

The follow-ups now fix every earlier correctness finding: retained PI state on cursor pages, live-row eviction, stale live-state precedence, pagination termination after parse failures, shared parser ordering, and empty client-filtered batch continuity. I verified the final simplification in `098adaf` preserves those fixes and corrects the merge/filter explanation. The latest completed CodeRabbit pass added only hook-order/constant-name nits, so I did not duplicate them.

## Backend and reuse research

Production accepts and enforces the generated one-column `id` cursor shape. [hyperlane-monorepo#9341](https://github.com/hyperlane-xyz/hyperlane-monorepo/pull/9341) is still open at `1be95e4`, so merge/deployment coordination remains necessary to make that support durable. I checked the adjacent SDK, widgets, and utils sources: there is no existing pagination/cursor primitive, but widgets already exports `Button`; that reuse is noted inline.

## Validation

- `pnpm test --runInBand`: 23 suites / 196 tests passed
- `pnpm run typecheck`: passed
- `pnpm run lint`: passed
- `git diff --check` against the PR merge base: passed
- Exact-head build, format, lint, typecheck, security, Socket, and Vercel checks: passed
- Live read-only GraphQL checks confirmed cursor direction/validation, the Katana sparse-pending path, and the oversized-bigint failure

No wallet, injection, secret, or dependency issues found.

src/features/messages/MessageSearch.tsx
troykessler: Fixed in 275d801. Cursor parsing now accepts only positive signed 64-bit IDs, so oversized values never reach PostgreSQL; invalid URL params are normalized back to the first page. Added max/zero/overflow regression coverage.

src/features/messages/MessageSearch.tsx
troykessler: Addressed in 275d801. Pagination controls now reuse the widgets Button primitive and share one pagination-specific class string, removing the repeated native-button styling.

paulbalaji (approved):
Approved exact head `c21e2ee7989a631f9db06d955d3a032bcefaf465`. All eight review threads are resolved and the fixes check out: raw cursor continuation survives filtering/parse failures, live state wins before filtering, empty filtered pages retain navigation, shared parser ordering is preserved, PI state stays first-page-only, and cursor inputs are bounded to the backend signed-bigint range. The final merge-main commit preserves the reviewed six-file patch exactly.

Validation: focused Jest 38/38, full Jest 23 suites / 199 tests, TypeScript, Oxlint, and `git diff --check` passed locally. Build, format, lint, typecheck, security, Socket, and Vercel checks pass; CodeRabbit remained queued at submission.

Xaroz (approved):
## Summary

Approved exact head `c21e2ee7989a631f9db06d955d3a032bcefaf465`. No actionable findings remain after re-reading the full six-file patch, all eight resolved review threads, and the related backend cursor implementation.

## Regression and contract checks

- Cursor input is restricted to positive signed PostgreSQL `bigint` values before GraphQL construction.
- Forward and reverse traversal use matching ID order/cursor semantics; a live first-page → next → previous smoke test returned to the original IDs.
- Raw DB IDs preserve continuation across parse failures and sparse client-side pending/warp filtering.
- Live rows override stale query rows only where live merging is enabled; historical pages remain stable.
- The parser's existing timestamp ordering remains unchanged for other callers.
- Existing monorepo primitives are reused where applicable: `@hyperlane-xyz/widgets` Button and `@hyperlane-xyz/utils` address helpers. I found no further worthwhile refactor.

The malformed direct-API cursor concern raised on the related monorepo PR is not exposed by this UI because invalid and out-of-range URL cursors are rejected before querying.

## Validation

- `pnpm test --runInBand`: 23 suites, 199 tests passed
- `pnpm run typecheck`: passed
- `pnpm run lint`: passed
- Exact-head GitHub build, format, lint, typecheck, security, dependency, and Vercel checks: passed

PR #359, opened 2026-09-01
feat: add EVM message self-relay

Xaroz:
## Summary

Adds an EVM-only self-relay action for pending messages that appear technically deliverable.

- Reuses the Widgets EVM chain configuration and Warp UI's RainbowKit/Wagmi setup for MetaMask, Rainbow, injected wallets, and WalletConnect.
- Verifies the exact message and dispatch transaction, delivery state, EVM protocols, validator metadata readiness, and destination `Mailbox.process` gas estimation before showing the relay action.
- Rechecks eligibility immediately before signing; the connected wallet only pays destination gas and Hyperlane's normal ISM verification remains enforced.
- Patches the transitive `brace-expansion` advisory introduced by the relayer metadata dependency.


src/pages/api/self-relay.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    const dispatchReceipt = await multiProvider
+      .getProvider(originChainName)
+      .getTransactionReceipt(originTxHash);
+    if (!dispatchReceipt) {
+      return res.status(404).json({ status: 'error', error: 'Dispatch transaction not found' });
+    }
+
+    const message = HyperlaneCore.getDispatchedMessages(dispatchReceipt).find(
```
paulbalaji: `HyperlaneCore.getDispatchedMessages` parses all receipt logs with the Mailbox ABI, but its SDK helper does not filter on `log.address`. A contract on any configured origin can emit a Dispatch-shaped event containing a chosen message, and this endpoint then treats it as canonically dispatched and performs recipient/ISM/metadata work. Destination verification still prevents delivery, but this bypasses the endpoint's trust check and makes the resource-amplification paths cheaper. Please filter the receipt logs by `coreAddresses[originChainName].mailbox` before decoding, with a test that rejects the same event from a non-Mailbox emitter.

src/pages/api/self-relay.ts
```diff
@@ -0,0 +1,157 @@
...
+      });
+    }
+
+    const ismAddress = await core.getRecipientIsmAddress(message);
+    const ism = await new EvmIsmReader(
+      multiProvider,
+      destinationChainName,
+      undefined,
+      message,
+    ).deriveIsmConfig(ismAddress);
```
paulbalaji: This derives a recipient-controlled ISM graph before the relayer's later `maxDepth` applies. In SDK 40.0.0, aggregation/routing derivation recursively expands module addresses without a visited set or depth/module/RPC budget. A self-referential or wide aggregation ISM can therefore keep this public request issuing destination RPCs until an infrastructure timeout. Please add cycle detection plus explicit depth/fanout/call/deadline bounds (and endpoint concurrency/in-flight limits), then test a self-cycle, a deep chain, and wide fanout for fast bounded rejection.

src/pages/api/self-relay.ts
```diff
@@ -0,0 +1,157 @@
...
+    // Match the CLI self-relay workaround: the canonical Merkle tree hook is
+    // the source of the checkpoint used to build validator metadata.
+    const hook = { type: HookType.MERKLE_TREE, address: merkleTreeHook };
+    const { BaseMetadataBuilder, isMetadataBuildable } =
+      await import('@hyperlane-xyz/relayer/metadata');
+    const metadataResult = await new BaseMetadataBuilder(core).build({
```
paulbalaji: This deterministically fails for every direct or nested OffchainLookup ISM: relayer 1.1.40 eagerly calls `multiProvider.getSigner(destination)` before even GET-style URL handling, while `createRelayContext` installs providers but no signer. CCTP routers use OffchainLookup metadata, and the UI currently permits them, so users can be offered self-relay that always returns a generic 500. #247 avoided this by attaching the connected wallet signer. Safest immediate fix is an explicit unsupported 422 plus UI gating. If server support is added, do not simply add a signer: ISM-controlled URLs then become an SSRF boundary requiring strict HTTPS/DNS/private-IP/redirect/body/time/concurrency controls.

src/features/wallet/EvmWalletContext.tsx
```diff
@@ -0,0 +1,59 @@
...
+    client({ chain }) {
+      const transport = fallback(chain.rpcUrls.default.http.map((url) => http(url)));
+      return createClient({ chain, transport });
+    },
+  });
+}
+
+export function EvmWalletContext({ children }: PropsWithChildren) {
+  const multiProvider = useMultiProvider();
+  const wagmiConfig = useMemo(() => createWagmiConfig(multiProvider), [multiProvider]);
```
paulbalaji: `useMultiProvider()` starts with an empty instance, replaces it when registry metadata becomes ready, and replaces it again on later metadata changes. Each identity change creates a new Wagmi config/store, which can reset an active or auto-reconnected wallet. This exact issue was raised on #247 and its final branch upgraded from a fallback config exactly once after `useReadyMultiProvider()` became available. Please preserve config/account continuity across empty-to-ready and later metadata refreshes, with a connected-wallet lifecycle test.

src/features/selfRelay/SelfRelayButton.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const isBusy = step !== 'idle';
+  const prepareRequest: SelfRelayPrepareRequest = {
+    messageId: message.msgId,
+    originDomainId: message.originDomainId,
+    originTxHash: message.origin.hash,
+  };
+  const prepareQuery = useQuery({
+    queryKey: ['selfRelayEligibility', message.msgId, message.origin.hash],
+    queryFn: () => prepareSelfRelay(prepareRequest),
+    refetchInterval: (query) =>
```
paulbalaji, line 38: This interval continues for permanent failures. `prepareSelfRelay` throws on every non-2xx/error response, leaving `query.state.data` undefined, so 400/404/422/500 paths still rerun every 30s despite `retry: false`. Each run can repeat receipt, delivery, ISM, validator/S3 metadata, and gas-estimation work. Please retry only the known transient 409/not-ready state (with a bound/backoff), stop on deterministic errors, and consider starting expensive preparation only after user intent; add interval tests for 409 versus permanent errors.

src/features/selfRelay/SelfRelayButton.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+      setStep('signing');
+      const hash = await sendTransaction(wagmiConfig, {
+        chainId: result.destinationChainId,
+        to: result.mailboxAddress,
+        data: result.calldata,
+      });
+
+      setStep('confirming');
+      await waitForTransactionReceipt(wagmiConfig, {
```
paulbalaji: Viem resolves this waiter with the replacement receipt for repriced, replaced, and cancelled transactions; the replacement reason is only exposed through `onReplaced`. Wagmi throws for a reverted receipt, but a standard cancellation transaction to self succeeds, so this code proceeds to the `Message relayed` toast even though `Mailbox.process` never ran. Please reject `cancelled`/semantically different replacements via `onReplaced`, or confirm actual message delivery before announcing success. Add cancellation, repricing, different-replacement, and revert tests.

paulbalaji (commented):
Reviewed exact head `268da444bb20549af4508b22f8bbfca435777327` against base `3f91a45e03473d40b91a9fed71f1a78dc44e55d2`. I found several material correctness/availability issues, so I am leaving a non-blocking comment rather than approving.

The most important findings are inline: unbounded/cyclic recipient ISM derivation on a public endpoint, acceptance of Dispatch-shaped logs from non-Mailbox emitters, deterministic CCIP-read/CCTP failure after moving metadata construction server-side, Wagmi state replacement when chain metadata refreshes, permanent API errors polling expensive preparation forever, and cancelled replacement transactions being reported as successful relays.

Two additional lower-severity concerns: the server relay context caches registry metadata/addresses/providers for the full warm-process lifetime, and this implementation restores the relayer's default 10s nested-aggregation timeout while rebuilding the metadata builder/cache on every request. PR #247's final branch had concrete mitigations for both the Wagmi initialization problem and slow nested metadata; those are worth retaining even though its browser proxy architecture is no longer needed. It also attached the connected wallet signer, which explains why its CCIP path did not have #359's provider-only failure.

I reviewed the existing PR feedback. CodeRabbit's missing-WalletConnect-ID comment is confirmed: RainbowKit 2.2.11 throws synchronously for the configured `projectId: ''`, and because the provider is mounted in `_app`, an environment without the new value loses the whole client. #247 already had a useful injected-only fallback. CodeRabbit's `brace-expansion@1.1.11` note is valid repository debt but the same advisory/copy exists on the base. Socket's warned packages were also already transitively reachable on the base; the relayer adds another dependency path, but I found no new pnpm-audit advisory ID relative to the base.

Validation at this exact head: build, typecheck, lint, format, diff-check, and focused self-relay tests pass. The suite including `TransactionCard.test.tsx` still has one stale Kiichain halt expectation; that failure is already present on current `main` after #356 and is not introduced here. There is no handler/component/E2E test exercising actual metadata building, CCIP/CCTP, wallet lifecycle, transaction replacement, or polling semantics.

Independent coverage: native adversarial review confirmed the endpoint/SDK findings. The requested Codex GLM profile (`glm-5.3`, read-only) completed successfully and independently reproduced the wallet, polling, replacement-transaction, and cache issues. Two requested Kimi `agent swarm` runs stalled after worker launch and were terminated, so I am not attributing any finding to Kimi. I also checked #247's commits and review threads for relevant regressions and confirmed which older proxy/signer issues #359 already avoids.

Xaroz:
@claude review

src/pages/api/self-relay.ts
Xaroz: Fixed in 03d6fcd using the safe path: direct or nested OffchainLookup derivation is rejected explicitly as HTTP 422. The client only re-polls HTTP 409, so the UI settles on an unsupported message instead of repeatedly returning a generic 500. No server signer or ISM-controlled URL fetching was added.

src/pages/api/self-relay.ts
Xaroz: Fixed in 03d6fcd. Receipt logs are filtered to the configured origin Mailbox before the SDK parser runs. The new test accepts the canonical emitter and rejects an identical Dispatch event from another contract.

src/features/selfRelay/SelfRelayButton.tsx
Xaroz: Fixed in 03d6fcd. Only HTTP 409 metadata-not-ready responses re-poll, with bounded 30s then 60s backoff; 400/404/422/429/500 stop immediately. Tests cover the transient and permanent status families.

src/features/selfRelay/SelfRelayButton.tsx
Xaroz: Fixed in 03d6fcd. The waiter records onReplaced: repricing remains valid, while cancellation and semantically different replacements are rejected; reverted receipts are also rejected before the success toast. Tests cover original success, repricing, cancellation, different replacement, and revert.

src/features/wallet/EvmWalletContext.tsx
Xaroz: Fixed in edc2e90, following the final #247 pattern. The wallet starts with a fallback config, upgrades exactly once after useReadyMultiProvider becomes available, and ignores later provider identity refreshes. The self-relay action is gated until that upgrade, and the lifecycle test verifies the ready Wagmi config remains identical across later provider replacements.

pnpm-workspace.yaml
```diff
@@ -7,6 +7,7 @@ minimumReleaseAgeExclude:
   - oxlint
   - '@oxlint/*'
 overrides:
+  'brace-expansion@>=2.0.0 <2.1.4': 2.1.4
```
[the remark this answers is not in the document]
Xaroz: Fixed in f99f629. Added the 1.x override, regenerated the lockfile, and verified brace-expansion 1.1.11 no longer resolves; minimatch 3 now uses 1.1.18.

src/consts/config.ts
```diff
@@ -6,6 +6,7 @@ const apiUrl = process.env.NEXT_PUBLIC_API_URL || 'https://explorer-api.hyperlan
 const wsUrl = resolveWsUrl(process.env.NEXT_PUBLIC_WS_URL, apiUrl);
 const registryUrl = process.env.NEXT_PUBLIC_REGISTRY_URL || undefined;
 const registryBranch = process.env.NEXT_PUBLIC_REGISTRY_BRANCH || 'main';
+const walletConnectProjectId = process.env.NEXT_PUBLIC_WALLET_CONNECT_ID || '';
```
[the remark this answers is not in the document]
Xaroz: Fixed in edc2e90. Blank values normalize to undefined, the app uses an injected-only connector fallback when WalletConnect is not configured, and .env.example now documents the optional real project ID. Added coverage for the fallback.

src/pages/api/self-relay.ts
Xaroz: Fixed in 03d6fcd. ISM derivation now has cycle, depth (10), aggregation fanout (16), total module (64), concurrency (4), and deadline bounds. The endpoint also has a 25s deadline and two-request in-flight cap. Tests cover self-cycle, deep traversal, wide fanout, total modules, timeout, and concurrency.

Xaroz:
## Review follow-up

Pushed three focused commits:

- `edc2e90` — stabilizes wallet initialization and handles a missing WalletConnect project ID.
- `03d6fcd` — hardens message validation, ISM traversal, API resource bounds, polling, and transaction replacement handling.
- `f99f629` — patches the remaining `brace-expansion` 1.x advisory.

Additional lower-severity items addressed:

- Relay registry/providers now refresh on a five-minute TTL instead of living for the full process lifetime.
- The lazily imported metadata builder is cached within that TTL context instead of being rebuilt per request.
- `useModal()` now stays with the component hook block.

Two suggestions were intentionally left unchanged:

- I did not carry over #247's 60-second nested aggregation timeout. This server endpoint has a hard 25-second request budget, so a longer child timeout would outlive the response. The in-flight slot remains occupied until underlying work settles, preventing timed-out work from bypassing the concurrency cap.
- `estimateGas` remains without `from`: preparation can run before a wallet is connected, a client-supplied address would be unauthenticated, and `Mailbox.process` is not sender-gated. The eventual wallet transaction is estimated/sent with the real connected account.

Preparation still starts to determine whether the message is self-relayable and to allow validator signatures to become ready later, but only the known transient `409` state now rechecks, twice with backoff. Permanent and unsupported states stop.

Validation: typecheck, lint, production build, and 25 focused tests pass. The full suite is 220/221; the only failure is the pre-existing Kiichain halt expectation already reproduced on `main`. A focused security pass found no remaining high/medium issue in this review diff.

Xaroz:
@claude review again

Pushed three focused commits:

- `edc2e90` — stabilizes wallet initialization and handles a missing WalletConnect project ID.
- `03d6fcd` — hardens message validation, ISM traversal, API resource bounds, polling, and transaction replacement handling.
- `f99f629` — patches the remaining `brace-expansion` 1.x advisory.

Additional lower-severity items addressed:

- Relay registry/providers now refresh on a five-minute TTL instead of living for the full process lifetime.
- The lazily imported metadata builder is cached within that TTL context instead of being rebuilt per request.
- `useModal()` now stays with the component hook block.

Two suggestions were intentionally left unchanged:

- I did not carry over #247's 60-second nested aggregation timeout. This server endpoint has a hard 25-second request budget, so a longer child timeout would outlive the response. The in-flight slot remains occupied until underlying work settles, preventing timed-out work from bypassing the concurrency cap.
- `estimateGas` remains without `from`: preparation can run before a wallet is connected, a client-supplied address would be unauthenticated, and `Mailbox.process` is not sender-gated. The eventual wallet transaction is estimated/sent with the real connected account.

Preparation still starts to determine whether the message is self-relayable and to allow validator signatures to become ready later, but only the known transient `409` state now rechecks, twice with backoff. Permanent and unsupported states stop.

Validation: typecheck, lint, production build, and 25 focused tests pass. The full suite is 220/221; the only failure is the pre-existing Kiichain halt expectation already reproduced on `main`. A focused security pass found no remaining high/medium issue in this review diff.

paulbalaji (commented):
Comprehensive rereview of exact head `f99f62916e28be84e2f91ed289ab75277c4b72d4`, including the existing reviews, replies, and threads. The follow-ups resolve the prior Mailbox-emitter validation, graph cycle/depth/aggregation-fanout/module bounds, OffchainLookup, Wagmi lifecycle, transaction replacement, cache, and brace-expansion findings. I still found four current-head issues below: one high-severity availability gap and three medium correctness/availability gaps.

Validation: frozen install, diff check, focused self-relay/wallet Jest (25/25), typecheck, lint, and production build passed. Full Jest was 220/221; the sole failure is the unchanged stale KiiChain halt expectation in `TransactionCard.test.tsx`. All current PR checks are green. Posting a non-blocking COMMENT as requested.

src/features/selfRelay/boundedIsmReader.ts
```diff
@@ -0,0 +1,126 @@
...
+    chain: string,
+    message: DispatchedMessage,
+    limits: IsmTraversalLimits,
+  ) {
+    super(multiProvider, chain, Math.min(limits.maxFanout, 4), message);
+    this.guard = new IsmTraversalGuard(limits);
+  }
+
+  override deriveIsmConfigFromAddress(address: Address): Promise<DerivedIsmConfig> {
+    return this.guard.visit(address, () => super.deriveIsmConfigFromAddress(address));
```
paulbalaji: **High — multisig validator fanout still bypasses these resource bounds.** `guard.visit` counts a multisig ISM as one module, but inherited `deriveMultisigConfig` accepts an arbitrary, duplicate-filled `validators` array. Relayer 1.1.40 passes every entry to `ValidatorAnnounce.getAnnouncedStorageLocations`, then uses unbounded `Promise.allSettled` fanout for S3 validator initialization and checkpoint GETs; its S3 helper also buffers each object without a byte cap. A recipient can return thousands of copies of one attacker-controlled announced validator, canonically dispatch one message, and make this public endpoint launch thousands of duplicate RPC/S3 operations while staying within `maxModules`/`maxFanout`. Please cap and dedupe validators before metadata construction, validate threshold/cardinality, and bound metadata-fetch concurrency/body size/deadline. The prior ISM-bounds thread is therefore only partially resolved.

src/features/selfRelay/serverLimits.ts
```diff
@@ -0,0 +1,32 @@
...
+    });
+  }
+}
+
+export class SelfRelayPreparationTimeoutError extends Error {}
+
+export async function withDeadline<T>(operation: Promise<T>, timeoutMs: number): Promise<T> {
+  let timeout: ReturnType<typeof setTimeout> | undefined;
+  try {
+    return await Promise.race([
```
paulbalaji: **Medium — the deadline does not terminate work or recover capacity.** This race returns a 503, but the original RPC/S3 promise keeps running; `InFlightLimit` decrements `active` only when that original promise settles. Two slow or never-settling operations therefore keep both per-instance slots occupied after their callers time out, making later requests return 429 indefinitely on that warm instance. The attacker-controlled S3 path above makes this externally influenceable. Please propagate an `AbortSignal` through preparatory I/O and add a non-settling-operation test proving the slot is eventually recovered after cancellation/cleanup.

src/features/selfRelay/SelfRelayButton.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const prepareRequest: SelfRelayPrepareRequest = {
+    messageId: message.msgId,
+    originDomainId: message.originDomainId,
+    originTxHash: message.origin.hash,
+  };
+  const prepareQuery = useQuery({
+    queryKey: ['selfRelayEligibility', message.msgId, message.origin.hash],
+    queryFn: () => prepareSelfRelay(prepareRequest),
+    refetchInterval: (query) =>
+      getSelfRelayRefetchInterval(query.state.error, query.state.fetchFailureCount),
```
paulbalaji: **Medium — repeated 409s still poll forever.** TanStack Query resets `fetchFailureCount` to 0 at the start of every fetch and increments it to 1 when that fetch fails. With `retry: false`, each interval-triggered refetch is a new fetch cycle, so this callback sees 1 after every 409 and never reaches the helper's stop-at-3 branch. I reproduced this with a browser-mode `QueryObserver`: counts alternated 0/1 while requests continued. Track metadata-pending attempts separately (or model 409 as successful pending data), and add an integration test around Query state transitions; the helper-only test cannot catch this.

src/features/selfRelay/boundedIsmReader.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    return {
+      address,
+      type: this.isZkSyncChain ? IsmType.STORAGE_AGGREGATION : IsmType.AGGREGATION,
+      modules: moduleConfigs,
+      threshold,
+    };
+  }
+
+  override async deriveOffchainLookupConfig(
```
paulbalaji, line 156: **Medium — OffchainLookup is not the only ISM this provider-only endpoint cannot build.** Relayer 1.1.40's `TRUSTED_RELAYER` builder calls `MultiProvider.getSignerAddress`, while `createRelayContext` installs no signers; `ARB_L2_TO_L1` calls `getSigner` and expects an Arb hook, while the endpoint always supplies a Merkle-tree hook. Both direct or routed forms therefore deterministically become generic 500s (and trusted-relayer configs exist in the registry). Please define the supported ISM/hook capability set and reject unsupported direct/nested nodes with the same stable 422 contract used here.

src/features/selfRelay/boundedIsmReader.ts
Xaroz: Confirmed and fixed in `bdf022a`.

Validator arrays are now rejected above 16 entries before metadata construction, deduplicated case-insensitively, and checked so the threshold is valid against the unique validator count. The request-scoped metadata adapter limits combined announcement/S3 work to 8 concurrent operations, restricts storage targets to validated AWS S3 HTTPS URLs, refuses redirects, streams each object with a 128 KiB cap, and propagates the request deadline signal. Tests cover duplicate/capped validator sets, threshold cardinality, concurrent fetch bounds, queued cancellation, unsafe targets, and advertised/streamed body limits.

src/features/selfRelay/serverLimits.ts
Xaroz: Confirmed and refined in `bdf022a` and `3b0ebf4`.

The deadline still returns a `503` promptly and propagates cancellation to prevent follow-on RPC/S3 work. The limiter now owns the underlying preparation promise independently, and `abortAfterSettled` keeps its capacity lease while an already-active, non-cancellable provider operation remains in flight. A regression test proves a timed-out pending operation continues to block another request until the original work settles.

src/features/selfRelay/SelfRelayButton.tsx
Xaroz: Confirmed and refined in `f1a7985` and `3b0ebf4`.

Only HTTP `409` metadata-not-ready responses are represented as successful `pending` query data. The persistent `dataUpdateCount` tracks attempts; polling runs after 30s and 60s, then stops after the third total pending result. `429`, `503`, and permanent errors surface immediately. Focus/reconnect refetches are disabled, and a browser-mode `QueryObserver` test proves no fourth request occurs.

src/features/selfRelay/boundedIsmReader.ts
Xaroz: Confirmed and refined in `bdf022a` and `caeddc6`.

Self-relay uses an explicit recursive capability check. Direct unsupported modules are rejected, and aggregations are accepted only when supported children can satisfy the configured threshold. This matters for the production Ethereum ISM used by the reported message: its nested 1-of-2 aggregation can build metadata through the supported message-ID multisig branch even though its optional Merkle-root branch is unsupported. If an unsupported branch is required for quorum, preparation still returns the stable `422` response. Tests cover both threshold outcomes.

src/pages/api/self-relay.ts
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  const parsedRequest = SelfRelayPrepareRequestSchema.safeParse(req.body);
+  if (!parsedRequest.success) {
+    return res.status(400).json({ status: 'error', error: 'Invalid self-relay request' });
+  }
+
+  const preparation = preparationLimit.run(() =>
+    withDeadline((signal) => prepareSelfRelay(parsedRequest.data, signal), PREPARATION_TIMEOUT_MS),
+  );
```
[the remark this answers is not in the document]
Xaroz: Confirmed and fixed in `3b0ebf4`.

The limiter now owns the underlying `prepareSelfRelay` promise, while the request deadline races it independently. `abortAfterSettled` prevents follow-on work after cancellation but keeps the capacity lease until an already-active, non-cancellable RPC operation actually settles. The regression test times out a still-pending operation, verifies another request remains blocked, then proves capacity is released only after the original work settles.

src/features/selfRelay/serverMetadata.ts
```diff
@@ -0,0 +1,294 @@
...
+      const validatorAnnounce = this.core.getContracts(originChain).validatorAnnounce;
+      const storageLocations = await this.ioLimit.run(
+        () => validatorAnnounce.getAnnouncedStorageLocations(toFetch),
+        this.signal,
+      );
+
+      await Promise.all(
+        toFetch.map(async (validator, index) => {
+          const storageLocation = storageLocations[index].at(-1);
+          if (!storageLocation) return;
```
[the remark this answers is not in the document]
Xaroz: Confirmed and fixed in `3b0ebf4`.

The paired lookup now uses optional access, so a short or malformed contract response is treated as a missing location for that validator instead of throwing before the per-validator handling and rejecting the full batch.

src/features/selfRelay/polling.ts
```diff
@@ -0,0 +1,27 @@
+import type { SelfRelayPrepareResponse } from './types';
+
+const MAX_PENDING_ATTEMPTS = 3;
+const PENDING_INTERVAL_MS = 30_000;
+const RETRYABLE_STATUS_CODES = new Set([409, 429, 503]);
```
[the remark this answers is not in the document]
Xaroz: Agreed and fixed in `3b0ebf4`.

Only HTTP `409` is represented as metadata-pending query data and receives the bounded 30s/60s polling sequence. `429` and `503` now surface immediately without automatic preparation retries; tests cover both groups.

paulbalaji (approved):
Comprehensive rereview of exact head `3b0ebf4` against base `3f91a45`, including the cumulative diff, all five follow-up commits since `f99f629`, and every existing review, reply, and thread. All 15 threads are resolved, and I found no remaining blocking or non-blocking correctness/security issue.

The follow-ups resolve the prior validator/S3 fanout, timeout-capacity, unbounded 409 polling, and unsupported-ISM findings. I also traced the bounded ISM/metadata path through the pinned SDK/relayer implementations and compared current production audit IDs with the exact base lockfile; this PR introduces no new advisory ID.

Validation: diff-check, 37 focused tests, typecheck, lint, format, and production build pass. Full Jest is 246/247; the sole failure is the unchanged stale KiiChain halt expectation previously reproduced on `main`, while the new underfunded-message test passes. All current GitHub checks are green.

Independent coverage: the requested Kimi agent swarm completed four review roles; its leads were rechecked and yielded no current defect. The Codex GLM lane performed an extensive read-only trace but did not converge to a final summary within the timebox, so I did not count it as a verdict or rely on it for approval.

